PR #5, opened 2016-02-10
New implementation of paml document pane, to complete replicate the xaml document pane.

abdelkarim:


grokys:
Is this PR ready to merge?

abdelkarim:
@grokys yes, the initial goal that was set, I believe is met, I am currently working on another PR that will allow the user disable the designer and fall back to just the editor in case someone is having issues with the designer.

PR #16, opened 2016-02-16
Moved Perspex.Designer from Perspex to PerspexVS

ImaBrokeDude:
* Part 1 Move designer to PerspexVS


ImaBrokeDude:
As a result of the intended implementation of the designer, it is not currently possible to move the designer to this repository.

abdelkarim:
@ImaBrokeDude can you please elaborate why is this not possible ?

ImaBrokeDude:
The designer uses an OWIN like approach in that when it is compiled some source files from other Perspexs libraries or integrated into the designer at build time. This allows the designer to handle displaying applications using different versions of Perspex within itself. Including the needed libraries and referencing them would break that functionally.

grokys:
I think it should be fine to just duplicate the shared files in PerspexVS - the whole idea of the OWIN-like system is to allow different versions so it shouldn't cause a problem.

I never liked Perspex.Designer being in Perspex itself.

ImaBrokeDude:
Okay, that fair actually haha. I will go ahead and do the migration PR.

PR #22, opened 2016-02-18
Improved the xaml editor implementation of the PerspexDesigner.

abdelkarim:
Previously the **WPF Editor** is extracted from the **IVsCodeWindow** instance that we create, the limitation of this approach was that we had to provide our own implementation for the various interfaces to have a good integration with Visual Studio. Some of which requires access to internal API, or I am not aware of how to achieve them at the moment. One of those is the ability to split the editor, and have two views of the same document.

So the solution was to wrap the **IVsCodeWindow** in our custom pane, and route all interface implementations to this instance, only one call failed which is the **OnSave(ref ...)**. the exception is raised in one of the internal classes, so I skipped its call by returning the success code.

Some interfaces are not yet implemented, those resides in **VsComInterfaces** class.

### DIFF
--- a/src/PerspexVS/Helpers/Extensions.cs
+++ b/src/PerspexVS/Helpers/Extensions.cs
@@ -1,6 +1,9 @@
 ﻿using Microsoft.VisualStudio.Settings;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Settings;
+using Microsoft.VisualStudio.Text;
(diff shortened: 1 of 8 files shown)


ImaBrokeDude:
Give me a little more time to test! This is very nice commit no less!

ImaBrokeDude:
Looks good to me. Works well. Not sure what the one error is able but I feel like it's related tot he custom tool :D. Nonetheless I think it's good so I'm gonna go ahead and merge it.

PR #35, opened 2016-07-08
Completition for Markup Extensions and minor fixes

donandren:
- code completition for Markup extension and Markup properties for:
  Static, Type, Binding (Converter and ConverterParameter}
- while solution is loading some project are null so better check for null instead throw exception
- suspend designer when Source view only is active


grokys:
@donandren I've been hoping someone else who knows the designer better than me could take a look at this. Is it ready for review? I seem to remember you saying there were problems?

donandren:
superseded by #76

PR #43, opened 2017-03-26
Sandbox: fix dnlib reference path

ForNeVeR:
It looks like someone forgot to change this path after Perspex being renamed back to Avalonia.


ForNeVeR:
Doesn't seems to be relevant with the latest changes; you fixed it yourself.

PR #56, opened 2018-01-14
Fix Avalonia.Ide submodule URL

grokys:
It was previously pointing to a non-existent submodule.


grokys:
Actually, is this submodule even used any more?

PR #81, opened 2019-04-05
Some fixes to templates

grokys:
The 0.8.0.0 release had a couple of problems in its templates:

- Was targeting `netcoreapp2.0` which is EOL. Target `netcoreapp2.1` instead which is a LTS
- The MVVM application template didn't create the `MainWindowViewModel` at runtime

### DIFF
--- a/templates/AvaloniaMvvmApplicationTemplate/Program.cs
+++ b/templates/AvaloniaMvvmApplicationTemplate/Program.cs
@@ -23,7 +23,12 @@ namespace $safeprojectname$
         // container, etc.
         private static void AppMain(Application app, string[] args)
         {
-            app.Run(new MainWindow());
+            var window = new MainWindow
+            {
+                DataContext = new MainWindowViewModel(),
+            };
+
+            app.Run(window);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


templates/AvaloniaMvvmApplicationTemplate/Program.cs
```diff
@@ -23,7 +23,12 @@ public static AppBuilder BuildAvaloniaApp()
         // container, etc.
         private static void AppMain(Application app, string[] args)
         {
-            app.Run(new MainWindow());
+            var window = new MainWindow
+            {
+                DataContext = new MainWindowViewModel(),
+            };
```
Kaev: This is the wrong approach imo.
I made a PR with the intended approach of the MVVM example. :)

grokys: Could you explain why you think this is the wrong approach? The approach that you use in your PR means that the designer will potentially instantiate a full runtime view model. The idea of `Design.DataContext` is that you can use sample data for prototyping rather than a full runtime instance of your application. See https://docs.microsoft.com/en-us/windows/uwp/data-binding/displaying-data-in-the-designer

Kaev: Now i understand why you used Design.DataContext! That wasn't clear to me.
It is really confusing for new users if the data binding works in the designer but not when you run the application. I was searching for any data binding errors when i first started to work with the MVVM template.
I think it's the wrong approach because in nearly all MVVM projects and tutorials i worked with the actual binding was made in the XAML code rather than the code behind - The code behind file wasn't even needed in most cases.

PR #83, opened 2019-04-09
Fixed data binding issue in MVVM template

Kaev:
Use Window.DataContext instead of Design.DataContext for proper data binding.


grokys:
As mentioned in https://github.com/AvaloniaUI/AvaloniaVS/pull/81, this will cause the full application to run in the designer, which is often not desired. I've merged #81 which will address the underlying issue so closing this.

Issue #104, opened 2019-07-26
Add "restart previewer process" button


grokys:
For when the previewer crashes.

kekekeks:
Also it would make sense to automatically restart the previewer when user is typing.

Issue #98, opened 2019-05-06
How to Disable Visual Previewer


Guddiny:
How to Disable Visual Previewer?

Issue #60, opened 2018-01-17
Intellisence not working with designer disabled


RexTremendae:
**Steps to Reproduce:**
* Disable the Avalonia designer in options
* Restart VS (not sure if this is needed for reproduction, but it might be)
* Open XAML file

**Expected Result:**
Intellisence should work even with the designer disabled.

**Actual Result:**
No intellisence.

Issue #54, opened 2017-11-02
Avalonia designer prevents copying the output file on build if it is a DLL


OronDF343:
This issue slows down development with Avalonia greatly for me

The Avalonia designer occasionally somehow prevents the build process from copying the output file. Once I start getting the issue it goes away if I wait a bit, and it is likely to happen on the next build as well.
It is the following error:

```
error MSB3027: Could not copy "<project dir>\bin\Debug\netstandard2.0\<project>.UI.dll" to "..\bin\Debug_Windows\<project>.UI.dll". Exceeded retry count of 10. Failed.
error MSB3021: Unable to copy file "<project dir>\bin\Debug\netstandard2.0\<project>.UI.dll" to "..\bin\Debug_Windows\<project>.UI.dll". The requested operation cannot be performed on a file with a user-mapped section open.
```

The DLL above is the one which contains the code for all of the UI components and targets .NET Standard 2.0, and the main EXE targets .NET Framework 4.6.1 and just initializes the UI.

I have not experienced this issue with any other project (none of them use Avalonia).

grokys:
@OronDF343 does the new version (0.8.0) work better in this respect?

grokys:
Thanks. When you do a build do you get get yellow banner at the top of the designer saying "the designer is paused because a build is in progress?"

OronDF343:
Yes, only when the designer is visible. If it is in a background tab I
don't see that, and am more likely to get the issue.

grokys:
Ok, could you look in the VS output window and show output from "Avalonia Diagnostics". You should see something like this appear when you start a build:

```
18:50:41.680 [Debug]  Build started
18:50:41.683 [Debug]  Setting pause state to True
18:50:41.686 [Verbose] 26144 Started PreviewerProcess.Stop()
18:50:41.688 [Information] 26144 Stopping previewer process
18:50:41.749 [Debug] 26144 Killing previewer process
18:50:41.752 [Information]  Process exited
18:50:41.756 [Verbose]  Started PreviewerProcess.Stop()
18:50:41.758 [Information]  Stopping previewer process
18:50:41.762 [Verbose]  Finished PreviewerProcess.Stop()
18:50:41.752 [Verbose]  Finished PreviewerProcess.Stop()
18:50:43.693 [Debug]  Build finished
18:50:43.695 [Debug]  Setting pause state to False
[...]

PR #107, opened 2019-10-19
designer improvements

donandren:
- attempt to fix lock of build output assemblies (used for preview and IntelliSense), by copying them to temp folder
- added option to disable preview, so only IntelliSense is active
- updated IntelliSense  system to be better supporting avalonia 0.9 features like avares compiled resources

available for early feedback: https://drive.google.com/open?id=1IVyIcFTUA17LMZ_QFr-_JCGepgyG3pwb

- partially solves the issue #104 (Enable/Disable should restart designer)
- #98 Fixed
- #60 Fixed
- #54 Fixed


grokys:
Thanks @donandren - do you think it'd be possible to split this up into separate PRs for easy review? In particular it would be nice to have the "copy to temp folder" feature as a separate PR so that we can discuss it separately, but also it makes it easier to review.

donandren:
The PR is more or less complete and tested, the extension is working decent, stable and with major improvements after this pr changes, it wouldn't make sense for me to put some additional work to split only the copy to temp functionality (it's really a small part of the PR a commit ), as it might take more time to do that instead fix or improve something else in my free time for avalonia (extracting a commit(s) then later dealing with conflicts can be a time eater :( don't want to risk broke what is already working pretty good)

grokys:
Ok, no problem - thanks for the work you've done on it!

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/AvaloniaUI/Avalonia
 [submodule "submodules/Avalonia.Ide"]
 	path = submodules/Avalonia.Ide
-	url = https://github.com/grokys/Avalonia.Ide
+	url = https://github.com/donandren/Avalonia.Ide
```
danwalmsley: We need to accept your changes into Avalonia.Ide

danwalmsley: prob should be: https://github.com/kekekeks/Avalonia.ide

@grokys @kekekeks maybe we need to get his repo moved to Avalonia organization?

donandren: that's correct. there is still outstanding pr  https://github.com/kekekeks/Avalonia.Ide/pull/22 which is not merged, then submodule can be forwarded to the original

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -185,11 +222,19 @@ public void Dispose()
...
+            var assemblyPath = SelectedTarget?.XamlAssembly;
+            var executablePath = SelectedTarget?.ExecutableAssembly;
+
             Process.FrameReceived -= FrameReceived;
 
             _throttle.Dispose();
             previewer.Dispose();
             Process.Dispose();
+
+            if (!alreadyDisposed && assemblyPath != null && executablePath != null)
```
kekekeks: This doesn't seem reliable

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -284,56 +331,156 @@ where IsValidOutput(output)
...
+            var executableDirDesigner = Path.Combine(designerTempDir, "exe");
+            var assemblyDirDesigner = Path.Combine(designerTempDir, "asm");
 
-            var assemblyPath = SelectedTarget?.XamlAssembly;
-            var executablePath = SelectedTarget?.ExecutableAssembly;
+            return (executableDirDesigner, assemblyDirDesigner);
+        }
 
-            if (assemblyPath != null && executablePath != null)
+        private (string executablePath, string assemblyPath) TryPrepareDesignTempData(string executablePath, string assemblyPath)
```
kekekeks: This code doesn't clean the previous data from "shadow copy" directory after re-build which potentially breaks user's app

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -284,56 +331,156 @@ where IsValidOutput(output)
             Log.Logger.Verbose("Finished AvaloniaDesigner.LoadTargetsAsync()");
         }
 
-        private async Task StartProcessAsync()
+        static private (string executableDir, string assemblyDir) GetDesignTempDirs(string executablePath, string assemblyPath)
         {
-            Log.Logger.Verbose("Started AvaloniaDesigner.StartProcessAsync()");
+            //let's try use very short folder names as limit for directory path is 248 chars
+            var designerTempDir = Path.Combine(Directory.GetParent(Path.GetDirectoryName(executablePath)).FullName, "dttmp");
```
kekekeks: This should at least be configurable by MSBuild properties.

One of the problems with shadow copying assemblies is that previewer effectively *runs the entire user's app*, which might need some data from other solution folders. The user code might have trouble finding those if we copy files to some random location.

Another problem is that people sometimes change their project output directory or just expect it not to have some random files laying around.

donandren: I can't see a potential problem here, as designer temp folder logic is chosen with that in mind.
if output folder of an app is:
`....\ApplicationOutput\bin\Debug\netcoreapp2.0` 
then designer temp folder will be:
`....\ApplicationOutput\bin\Debug\dttmp` 
so all the relative and absolute paths from the app if any dependencies should work fine.
Agree user might not like copy files to a random location, but not sure we can fix build output locks with different approach than copy output to a temp folder :(
we can make attempt to have settings in designer or ms build property where and what should be the name of the temp folder, but doubt anybody will use that option.

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
donandren: binaries in output folder are not deleted on purpose, they are overridden with new files only if changed, so  we don't lock ourselves while lunching the designer process. Even if something wrong happend (unpredictable), shouldn't be a big issue as designer will clean up the temp folder when closed

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
donandren: dispose might be called twice, but haven't been able to understand why

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
kekekeks: There might be files that shouldn't exist anymore after rebuild.

.gitmodules
grokys: You shouldn't need to change the url of the submodule, the commit will be available from the original URL even if it was made in a fork.

grokys:
When I get time later this week, I'll try to split this PR up into its component parts. @donandren in future if you could open separate PRs for each feature/bugfix it would be really helpful:

1. It makes it easier to review; otherwise:
  - It's hard to know what change relates to which fix/feature
  - One feature/fix can hold up the other featrues/fixes in the PR (such as "copy output directory")
2. It makes it easier to understand why a change was made in future (I've been needing this a lot recently!)
3. It makes it easier to revert a change should it cause problems

AvaloniaVS/Services/IAvaloniaVSSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.ComponentModel;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Serilog.Events;
+
+namespace AvaloniaVS.Services
+{
+    public enum DesignerViewType
+    {
+        Split,
+        Design,
```
grokys: What is the use-case for just displaying the preview without the XAML?

grokys: Ahh, never mind - it's the default view, it can still be switched ;)

grokys:
Unchecking "Preview Enabled" seems to cause endless exceptions to be thrown:

image

AvaloniaVS/AvaloniaPackage.cs
```diff
@@ -15,10 +17,12 @@ namespace AvaloniaVS
     [InstalledProductRegistration("#110", "#112", "1.0", IconResourceID = 400)]
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [ProvideAutoLoad(VSConstants.UICONTEXT.SolutionExistsAndFullyLoaded_string, PackageAutoLoadFlags.BackgroundLoad)]
-    [ProvideEditorExtension(typeof(EditorFactory), ".paml", 100, NameResourceID = 113, DefaultName = "Avalonia Xaml Editor")]
+    [ProvideEditorExtension(typeof(EditorFactory), ".paml", 100, NameResourceID = 113, EditorFactoryNotify = true, ProjectGuid = VSConstants.UICONTEXT.CSharpProject_string, DefaultName = Name)]
+    [ProvideEditorExtension(typeof(EditorFactory), ".xaml", 0x40, NameResourceID = 113, EditorFactoryNotify = true, ProjectGuid = VSConstants.UICONTEXT.CSharpProject_string, DefaultName = Name)]
```
grokys: Why were these two changes necessary? What issue/feature/bugfix do they relate to?

AvaloniaVS/Views/AvaloniaDesigner.xaml
```diff
@@ -11,55 +11,75 @@
...
 
         <DockPanel Name="toolbar"
                    Grid.Row="0"
                    Margin="4"
                    VerticalAlignment="Top">
-            <ComboBox DockPanel.Dock="Right"
+            <StackPanel Orientation="Horizontal" DockPanel.Dock="Right">
+                <CheckBox Content="Preview Enabled" IsChecked="{Binding IsPreviewEnabled, ElementName=root, Mode=TwoWay}" Margin="0 0 10 0" 
+                          VerticalAlignment="Center" local:VsTheme.UseVsTheme="True"/>
+                <ListBox Margin="0 0 10 0" BorderThickness="0"
```
grokys: I wonder if we need to put this in the options dialog? My initial thought is that the designer should just remember the last mode used when opening an new designer.

AvaloniaVS/AvaloniaPackage.cs
grokys: The commit says "register extension in open with context menu" but I'm not sure what that means...

AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
```diff
@@ -127,51 +127,58 @@ private bool HandleSessionUpdate(char c)
...
+            }
+
+            var selected = _session.SelectedCompletionSet.SelectionStatus.Completion as XamlCompletion;
+
+            if (!_session.SelectedCompletionSet.SelectionStatus.IsSelected || selected == null)
+            {
+                return false;
+            }
+
+            if (c == '/')
```
grokys: What issue is this fixing? It seems to be making sure the element is closed when `/` is typed, but as far as I can tell that already worked.

AvaloniaVS/AvaloniaPackage.cs
donandren: image
option in dialog after `Open With...`

donandren: that's really useful for xaml files with non avalonia namespace base class

AvaloniaVS/Views/AvaloniaDesigner.xaml
donandren: it's different, we have settings which are persisted (in Settings) and one for per editor. 
my workflow is like that:
- disabled preview and using only code completion
- enable preview manually to check something in ui, then switch back to xaml only edit (don't want persisted in another editor instances/xaml files)

AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
donandren: that's a code which i don't have a clue , but didn't wanted to delete it as i don't understand it.
But tried to simplify the logic here and extract the unknown part as it was having wierd logic that was breaking completion of `avares://Assembly/Resource.xaml` like urls

AvaloniaVS/Services/IAvaloniaVSSettings.cs
donandren: designer is also xaml editor with completions, there is a use case to use it only for xaml code editing and no preview

.gitmodules
donandren: ha i didn't knew that

grokys:
Regarding intellisense, I see it working when filling in `x:Class` attributes, but I can't get it to work when doing e.g. `<Image Source="` - should it be popping up intellisense there?

grokys:
What is the use case for having both a "Preview Enabled" checkbox and a "Design View" option? When would you want to disable the previewer while viewing the design view and when would you want the previewer enabled when in the source view?

To my mind, the previewer should be started/stopped based on which view you're currently in. If the problem is #104 then IMO a button should appear when the previewer process stops to allow you to restart it.

donandren:
> What is the use case for having both a "Preview Enabled" checkbox and a "Design View" option? When would you want to disable the previewer while viewing the design view and when would you want the previewer enabled when in the source view?
> 
> To my mind, the previewer should be started/stopped based on which view you're currently in. If the problem is #104 then IMO a button should appear when the previewer process stops to allow you to restart it.

i think we can live without enable preview checkbox, when we have split/code/preview options, although it might be useful is some scenarios, like disable preview  until you type xaml without changing layout then enable it back for reference, wouldn't mind we drop it tho

donandren:
> Regarding intellisense, I see it working when filling in `x:Class` attributes, but I can't get it to work when doing e.g. `<Image Source="` - should it be popping up intellisense there?

for `<Image Source="`  onyl resm like resources are completed
`avares://` completion is working for xaml avalonia resources without x:Class only - for sure there is a room for improvements

donandren:
@grokys  i can start working on this pr notes after your pr's are merged

grokys:
@donandren all PRs merged. I'll close this PR now as there are a lot of conflicts.

donandren:
@grokys  a couple of major fixes hasn't been missed from this pr in the small pr's that has been made. I'll keep this open until i verify all important pieces are translated to another pr's

donandren:
> @donandren all PRs merged. I'll close this PR now as there are a lot of conflicts.

this was expected

PR #116, opened 2019-11-28
Enable quick find in editor.

### LINKED ISSUE #108: Quick find and navigating to previous file doesn't work with xaml editor
Visual Studio 2019 16.3.5

"Quick Find" launched by Ctrl + F opens "Find and Replace" instead of "Quick Find" when focus is in xaml code editor window
"Navigate Back" launched by Ctrl + - omits all previously opened *.xaml files and goes straight to nearest *.cs file

It doesn't work in other VS versions as well and it doesn't seem to be working in VS2017.

In WPF editor it works fine.

grokys:
Previously pressing Ctrl+F in the avalonia xaml editor would open the Find dialog instead of the Quick Find dialog. Fixed this by implementing an internal VS interface. Is there a better way to do this?

image

Fixes #108

### DIFF
--- a/AvaloniaVS/AvaloniaVS.csproj
+++ b/AvaloniaVS/AvaloniaVS.csproj
@@ -65,6 +65,7 @@
     <Compile Include="Models\ProjectInfo.cs" />
     <Compile Include="Models\ProjectOutputInfo.cs" />
     <Compile Include="Models\XamlBufferMetadata.cs" />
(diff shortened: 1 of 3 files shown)


grokys:
>  i wonder how you guys find this kind of information?

[dnspy](https://github.com/0xd4d/dnSpy) and a lot of patience!

PR #120, opened 2019-12-01
Don't ship Avalonia.Designer.HostApp.

### LINKED ISSUE #110: Don't ship Avalonia.Designer.HostApp.dll
Instead use the one pointed to by `AvaloniaPreviewerNetCoreToolPath`/`AvaloniaPreviewerNetFullToolPath`.

grokys:
Instead of shipping our own version of `Avalonia.Designer.HostApp`, inspect the MSBuild properties of the project we're previewing to find the correct version.

attn: @kekekeks 

Fixes #110


AvaloniaVS/Services/SolutionService.cs
```diff
@@ -203,68 +209,31 @@ private static async Task<IReadOnlyList<ProjectOutputInfo>> GetOutputInfoAsync(P
...
 
-            string fullPath = TryGetProperty(project?.Properties, "FullPath");
+                        var outputInfo = new ProjectOutputInfo
+                        {
+                            TargetAssembly = targetPath,
+                            TargetFramework = targetFw == "classic" ? "net40" : targetFw,
+                            HostApp = hostAppNetCore,
+                        };
 
-            string outputPath = project?.ConfigurationManager?.ActiveConfiguration?.Properties?.Item("OutputPath")?.Value?.ToString();
```
grokys: I removed this whole part here because it wasn't accessing MSBuild and I couldn't find a case where it was needed: every example I tried worked by inspecting the `UnconfiguredProject`. @kekekeks any idea of this code is in fact still needed? If so, how do I create a project that needs it.

kekekeks: I think it was needed for old SDK projects.

PR #125, opened 2019-12-10
Add optional shadow copy assemblies for designer

donandren:
added  a way to prevent lock of build output (which tend to happends very very often after 10-15 mins work with the extension).

The shadow copy option is configurable from the avalonia extension settings and be default is **off!**
The temp folder to copy application output for the designer run is also configurable.

When option enabled it just copy target application output to a configured temp folder and launches the app in a design mode.

extracted implementation from #107 with some improvements
Fixes: #54 

Note: this solution have been tested by me for several months and once i've started using this solution i didn't expirienced any lock of build output because of avalonia designer

### DIFF
--- a/AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
+++ b/AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
@@ -203,6 +203,8 @@ namespace AvaloniaVS.Views
         /// </summary>
         public void Dispose()
         {
+            var alreadyDisposed = _disposed;
+
             _disposed = true;
 
             if (_editor?.TextView.TextBuffer is ITextBuffer2 oldBuffer)
@@ -215,11 +217,19 @@ namespace AvaloniaVS.Views
                 _editor.Close();
             }
 
+            var assemblyPath = SelectedTarget?.XamlAssembly;
+            var executablePath = SelectedTarget?.ExecutableAssembly;
+
             Process.FrameReceived -= FrameReceived;
 
             _throttle.Dispose();
             previewer.Dispose();
             Process.Dispose();
+
+            if (!alreadyDisposed && assemblyPath != null && executablePath != null)
+            {
+                _ = Task.Delay(100).ContinueWith(t => TryCleanDesignTempData(executablePath, assemblyPath));
+            }
         }
 
         protected override void OnDpiChanged(DpiScale oldDpi, DpiScale newDpi)
@@ -361,6 +371,13 @@ namespace AvaloniaVS.Views
 
             if (assemblyPath != null && executablePath != null && hostAppPath != null)
             {
+                var d = TryPrepareDesignTempData(executablePath, assemblyPath);
+
+                if (!d.Equals(default((string, string))))
+                {
+                    assemblyPath = d.assemblyPath;
+                    executablePath = d.executablePath;
+                }
 
                 RebuildMetadata(assemblyPath, executablePath);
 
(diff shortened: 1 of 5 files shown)


AvaloniaVS/Views/OptionsView.xaml
```diff
@@ -40,5 +44,17 @@
...
+
+        <CheckBox x:Name="chkCopyOutput" Grid.Row="6" Grid.Column="0" Grid.ColumnSpan="3" IsChecked="{Binding UseTempDir, Mode=TwoWay}">
+            <StackPanel>
+                <TextBlock Text="Shadow copy application output to temp folder for designer" TextWrapping="Wrap" />
+                <TextBlock Text="(helps with lock of build output)" />
+            </StackPanel>
+        </CheckBox>
+
+        <TextBlock Grid.Row="8" IsEnabled="{Binding UseTempDir}" Text="Designer Temp Folder Path
+                   (Relative to $(OutputPath) or absolute)" TextWrapping="Wrap"/>
```
grokys: What is the use-case for allowing the user to specify the temp dir location?

Also won't absolute paths cause problems when opening XAML files from two different projects simultaneously?

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -429,6 +448,109 @@ private void RebuildMetadata(string assemblyPath, string executablePath)
             }
         }
 
+        private (string executableDir, string assemblyDir) GetDesignTempDirs(string executablePath, string assemblyPath)
+        {
+            if (string.IsNullOrEmpty(DesignerTempPath))
+                return default((string, string));
+
+            //let's try use very short folder names as limit for directory path is 248 chars
```
grokys: This comment isn't really true if you let the user specify the temp dir path...

AvaloniaVS/Views/OptionsView.xaml
donandren: use might want custom temp folder names/logic (also @kekekeks suggested one should be able to customise temp dir) 
so i would agree we better have this option with good defaults and at least it's visible and clear which folder will be used for temp one

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
donandren: haha yea, it's really annoying when you hit max path, happend to be several times during my dev expirience

AvaloniaVS/Views/OptionsView.xaml
grokys: Ok, should we allow absolute paths? Won't they cause problems when opening XAML files from two different projects simultaneously?

donandren: agree we better have only relative paths, anyway the way temp path is constructed expect relative path

grokys: Ok, we you going to add a check to make sure it's a relative path, or is that unnecessary do you think?

donandren: i don't think it's necessary to have a check. Temp folder name (or path) is something that i doubt anybody will have a need to change it, as default option should be good enough. Anyway the real validation is only when temp folder will be used in a specific project and hard to make sure without  a working solution path is really valid. But if you strongly disagree i can try add some simple validation.

grokys: No, I agree that validation probably isn' t necessary. I'll take a final look over this (it's been a while) and get it merged!

AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -429,6 +448,108 @@ private void RebuildMetadata(string assemblyPath, string executablePath)
             }
         }
 
+        private (string executableDir, string assemblyDir) GetDesignTempDirs(string executablePath, string assemblyPath)
+        {
+            if (string.IsNullOrEmpty(DesignerTempPath))
+                return default((string, string));
+
+            var designerTempDir = Path.Combine(Directory.GetParent(Path.GetDirectoryName(executablePath)).FullName, DesignerTempPath);
```
kekekeks: Does that mean that designer temp dir would be something like `ProjectName\bin\Debug\dttemp`?

kekekeks: In that case what will happen if user has multitargeting enabled? Will all of the target frameworks share the same designer path?

kekekeks: BTW, we could somewhat standardize designer temp path by introducing an MSBuild property from Avalonia package. That way the path could be customized on per-project basis. That would probably have to wait till 0.10 though.

akunchev: **Note:** copy to temp folder is optional and disabled by default
> Does that mean that designer temp dir would be something like `ProjectName\bin\Debug\dttemp`?
yes

> In that case what will happen if user has multitargeting enabled? Will all of the target frameworks share the same designer path?

designer works with only one preselected target and probably not an issue in 99.99%, not sure designer launches one app at a time or multiple? but in theory there might be a potential issue in a weird scenario

donandren: **Note:** copy to temp folder is optional and disabled by default
> Does that mean that designer temp dir would be something like `ProjectName\bin\Debug\dttemp`?

yes

> In that case what will happen if user has multitargeting enabled? Will all of the target frameworks share the same designer path?

designer works with only one preselected target and probably not an issue in 99.99%, not sure designer launches one app at a time or multiple? but in theory there might be a potential issue in a weird scenario

kekekeks: There might be two tabs with two different XAML files from the same project opened at the same time with different TFMs selected.

donandren:
no longer experiencing output locks with latest extension, no need of this pr

PR #126, opened 2019-12-11
port xaml completion improvements from #107

donandren:
- added support for completions of `avares://` 
- when completed attached property like Grid.Row completion is auto shown after `Grid.`
- if project has multiple target apps designer selection is preserved between builds
- current assembly name is passed to Avalonia.Ide infrastructure, so assembly local resources will be shown in the short form
- fixed code completion works when code only view is active (it was disabling before not only preview but also code completion)

**Note:** need submodule update for intelisense improvements

Dependent of Avalonia.Ide changes in pr: https://github.com/kekekeks/Avalonia.Ide/pull/22


grokys:
@donandren could you describe what these XAML completion improvements are?

donandren:
> @donandren could you describe what these XAML completion improvements are? #107 just says:
> 
> > updated IntelliSense system to be better supporting avalonia 0.9 features like avares compiled resources
> 
> But when I tried it, typing `avares:` didn't bring up intellisense. It would be good to get a description of what functionality this PR is supposed to enable.

just start typing (in a StyleInclude Source=" ) a should list `avares:/`/ resources, also local assembly avares will be shown in the sort form like `/Styles/MyStyle.xaml`

avares completion is done properly for styles.
For Images etc. it's not possible to be done only with reflection because of the internal design of avares and how it works, completion should work properly for regular assembly resources with `resm://` tho

donandren:
@grokys added support for all `avares://` resources now `<Image Source="` should be completed if  resources are built in assembly

PR #131, opened 2020-01-05
Change GridSplitter style to look more like the WPF designer's splitter

Fusion86:
The color is identical to the WPF designer's when using the Dark or Light theme.
It looks like they use a hard-coded value for the Blue theme, or maybe I just couldn't find the right color key. So when using the Blue theme it looks a bit darker than the WPF splitter. The same is true for the blue contrast theme.

I also made the splitter wider (just like in the WPF designer) because the smaller one was easier to overlook (user might not be aware that he can move it) and it's easier to grab. But I can revert that if wanted.


Fusion86:
Sure thing. [This is what it looks like now.](https://i.imgur.com/YXM73wU.png)

PR #165, opened 2020-07-08
Update using dpi info

rstm-sf:
Hello!

Update using dpi info. We get the current dpi after this changes https://github.com/AvaloniaUI/Avalonia/pull/3273 

Depends on Avalonia release 0.10?

### DIFF
--- a/AvaloniaVS/AvaloniaVS.csproj
+++ b/AvaloniaVS/AvaloniaVS.csproj
@@ -21,6 +21,7 @@
     <RootNamespace>AvaloniaVS</RootNamespace>
     <AssemblyName>AvaloniaVS</AssemblyName>
     <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
+    <LangVersion>8.0</LangVersion>
     <GeneratePkgDefFile>true</GeneratePkgDefFile>
     <IncludeAssemblyInVSIXContainer>true</IncludeAssemblyInVSIXContainer>
     <IncludeDebugSymbolsInVSIXContainer>true</IncludeDebugSymbolsInVSIXContainer>
--- a/AvaloniaVS/Services/PreviewerProcess.cs
+++ b/AvaloniaVS/Services/PreviewerProcess.cs
@@ -28,7 +28,7 @@ namespace AvaloniaVS.Services
         private readonly ILogger _log;
         private string _assemblyPath;
         private string _executablePath;
-        private double _scaling;
+        private Vector _dpi;
         private Process _process;
         private IAvaloniaRemoteTransportConnection _connection;
         private IDisposable _listener;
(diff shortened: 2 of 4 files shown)


AvaloniaVS/Services/PreviewerProcess.cs
```diff
@@ -400,8 +400,8 @@ private async Task OnMessageAsync(object message)
                         _bitmap = new WriteableBitmap(
                             Math.Max(frame.Width, 1),
                             Math.Max(frame.Height, 1),
-                            96,
-                            96,
+                            frame.DpiX,
+                            frame.DpiY,
```
rstm-sf, line 404: dpi is not always sent _(= 0)_, but in the end it turns out the one that is needed after some internal updates?

rstm-sf:
Can we exchange for any other image? https://github.com/AvaloniaUI/AvaloniaVS/blob/d4e36c9539a5a6f79085c3ece91e196b277478ba/azure-pipelines.yml#L7

vs2017 does not support language version 8.0. I understand correctly that [this](https://github.com/AvaloniaUI/Avalonia/blob/031e6cd1c43e334b67fa2daf996b660f016f31f1/src/Markup/Avalonia.Markup.Xaml/XamlIl/CompilerExtensions/XamlIlBindingPathHelper.cs#L256-L287) is a feature of version 8?

AvaloniaVS/AvaloniaVS.csproj
```diff
@@ -85,9 +86,7 @@
     <Compile Include="Views\AvaloniaDesigner.xaml.cs">
       <DependentUpon>AvaloniaDesigner.xaml</DependentUpon>
     </Compile>
-    <Compile Include="Views\OptionsDialogPage.cs">
-      <SubType>Component</SubType>
-    </Compile>
+    <Compile Include="Views\OptionsDialogPage.cs" />
```
rstm-sf, line 89: --

AvaloniaVS/Services/PreviewerProcess.cs
rstm-sf: Although I observe the resizing during rendering and in the release (as if it comes at the beginning with 96 dpi, but eventually adapts)

grokys:
> Can we exchange for any other image?

Yeah I think we should be using VS2019, I didn't realize we were still on VS2017. We should probably drop support for VS2017 too as I've heard reports it doesn't work well.

grokys:
> Depends on Avalonia release 0.10?

Why does this depend on 0.10?

rstm-sf:
> Why does this depend on 0.10?

These changes will come along with 0.10, so I thought that there was some kind of dependence. But it seems that due to the fact that it is connected as a submodule, there are no problems. For example, a project with 0.9.7 normally shows me. Sorry, I do not understand well

grokys:
> Can we exchange for any other image?

I tried updating the image to VS2019 in #166 and for some reason the CI hangs. Thinking back I seem to remember seeing this problem before which is why we're still on VS2017 :/ Not sure why this is happening or what the fix is.

grokys:
@rstm-sf I'm unclear on what the issue is that this PR is fixing, could you explain the problem the PR is solving?

rstm-sf:
@grokys, starting from 0.10, avalonia can send the dpi in which it was rendered, and we no longer need to change the images for scale.

I think that when I tried various versions, I overlooked something and I did it well, but according to the idea, this should not have happened ...

Since we have problems with CI and this change will hardly affect the user in any way, I think we can forget about it for now. Thanks!

AvaloniaVS/Services/PreviewerProcess.cs
rstm-sf: I think that I was mistaken here and this should not have been the Avalonia version 0.10. Therefore, it is better to check for zero

Issue #133, opened 2020-02-11
Intellisense missing all events


Symbai:
All kind of control events (e.g. `Scroll` on `Scrollbar`) are missing in intellisense. Makes it really difficult porting WPF code to Avalonia. Tested on AvaloniaVS 0.9.0.0 in VS 2019 Community Edition.

wdcossey:
R# also not working

ShadowDancer:
@grokys what do we need for this? Simply pushing all events provided Dnlib to Intellisense will be good starting point?

grokys:
Fixed by #167

PR #167, opened 2020-07-16
Add event completion

ShadowDancer:
## What does the pull request do?
Add intellisense completion for event and attached events

## What is the current behavior?
There is no completion for event or attached event

## What is the updated/expected behavior with this PR?
When typing name of event it is completed.
When typing name of type, which has attached events it is completed
When typing name of type, which has attached events + dot it will complete event name
Try typing event name ie. `KeyUp` on control, or `Gestures.*` for attached event

## How was the solution implemented (if it's not obvious)?
Events are straightforward. Dnlib does not differentiate between public and private events (idk if this is how framework works), so all are provided.

For attached properties I am looking for public static fields and (return type starts with `RoutedEvent` `or` name ends with `Event`).
I am not sure if `or` should be changed to `and`. Also maybe there is need for checking static properties?

## Fixed issues
#133

Also see PR on Avalonia.Ide
https://github.com/kekekeks/Avalonia.Ide/pull/26

### DIFF
--- a/AvaloniaVS/IntelliSense/XamlCompletion.cs
+++ b/AvaloniaVS/IntelliSense/XamlCompletion.cs
@@ -74,10 +74,12 @@ namespace AvaloniaVS.IntelliSense
 
             s_images = new ImageSource[capacity];
             s_images[(int)CompletionKind.Property] = LoadImage(imageService, KnownMonikers.Property, ref attributes);
+            s_images[(int)CompletionKind.Event] = LoadImage(imageService, KnownMonikers.Event, ref attributes);
             s_images[(int)CompletionKind.Class] = LoadImage(imageService, KnownMonikers.MarkupTag, ref attributes);
             s_images[(int)CompletionKind.Enum] = LoadImage(imageService, KnownMonikers.EnumerationItemPublic, ref attributes);
             s_images[(int)CompletionKind.Namespace] = LoadImage(imageService, KnownMonikers.Namespace, ref attributes);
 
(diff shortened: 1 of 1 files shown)


ShadowDancer:
@grokys can you comment on how to properly detect attached event?

As I understand only fields with `RoutedEvent<>` or `RoutedEvent` are interesting for us?
I've made it so that if return type is `RoutedEvent*`, or name ends with `Event` it will be treated like attached event.

grokys:
Hi @ShadowDancer - I'm back off vacation now. Taking a look.

grokys:
> As I understand only fields with RoutedEvent<> or RoutedEvent are interesting for us?
> I've made it so that if return type is RoutedEvent*, or name ends with Event it will be treated like attached event.

Yeah, I think that sounds like a reasonable approach. From my quick testing it finds the events I'd expect it to find. We can always tweak it later if necessary.

ShadowDancer:
I think it should work like this:
1. find associated .xaml.cs file
2. add method to codebehind class

Point 1. should be easy, for 2 I don't see easy way to add method to class without too much plumbing as there is so many corner cases (ie. class may be nested in multiple namespace bocks, etc.).

ShadowDancer:
Also can you merge these changes in submodule?

grokys:
> Also can you merge these changes in submodule?

I don't have write access to that repository, as it's one of @kekekeks personal projects so he'll have to review/merge.

ShadowDancer:
It's a bit weird that changes are merged in VS repository and not in Ide.

Issue #174, opened 2020-10-14
F# templates


FoggyFinder:
Would be awesome to have. Currently it's inconvenient especially when it's come to adding new UserControl/Window

grokys:
If someone wants to submit a PR with F# templates and someone else wants to review them, I'd be 👍 with including! Unfortunately I don't know F# well enough to feel confident doing this myself.

FoggyFinder:
Awesome. I know a very little about VS templates but it shouldn't be too complicated. I'll try to create PR on weekend

PR #175, opened 2020-10-16
Add F# templates

FoggyFinder:
Closes: #174 

Some parts were copied from C# projects, unsure whether it is really necessary to have.

For example

```xml
  <ItemGroup>
    <BootstrapperPackage Include=".NETFramework,Version=v4.6">
      <Visible>False</Visible>
      <ProductName>Microsoft .NET Framework 4.6 %28x86 and x64%29</ProductName>
      <Install>true</Install>
    </BootstrapperPackage>
    <BootstrapperPackage Include="Microsoft.Net.Framework.3.5.SP1">
      <Visible>False</Visible>
      <ProductName>.NET Framework 3.5 SP1</ProductName>
      <Install>false</Install>
    </BootstrapperPackage>
  </ItemGroup>
```

Also, noticed that icons for `User Control`, `Window` templates aren't displayed. Was it done purposely?

Anyway, ready for review and/or feedback


templates/FsAvaloniaWindowTemplate/Window.xaml.fs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Views
+
+open Avalonia
+open Avalonia.Controls
+open Avalonia.Markup.Xaml
+
+type $safeitemrootname$ () as self =
```
Kir-Antipov, line 7: `$safeitemrootname$` is equivalent to the `$safeitemname$`, but it's not available in pre-2017 versions that are still in use. So it will be better to replace this with the more accessible `$safeitemname$`.

templates/FsAvaloniaWindowTemplate/Window.xaml
```diff
@@ -0,0 +1,12 @@
+﻿<Window
+    x:Class="Views.$safeitemrootname$"
+    xmlns="https://github.com/avaloniaui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    Title="$safeitemrootname$"
```
Kir-Antipov, line 7: Same as for [templates/FsAvaloniaWindowTemplate/Window.xaml.fs](https://github.com/AvaloniaUI/AvaloniaVS/pull/175/files#diff-ae5a8606b1789a581e1e75b2cfe0aa7ff0428bde09fb9929406c9de2b5009a96R7).

templates/FsAvaloniaUserControlTemplate/UserControl.xaml.fs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Views
+
+open Avalonia
+open Avalonia.Controls
+open Avalonia.Markup.Xaml
+
+type $safeitemrootname$ () as self =
```
Kir-Antipov, line 7: `$safeitemrootname$` -> `$safeitemname$`

templates/FsAvaloniaUserControlTemplate/UserControl.xaml
```diff
@@ -0,0 +1,11 @@
+﻿<UserControl
+    x:Class="Views.$safeitemrootname$"
```
Kir-Antipov, line 2: `$safeitemrootname$` -> `$safeitemname$`

templates/FsAvaloniaMvvmApplicationTemplate/nuget.config
```diff
@@ -0,0 +1,11 @@
...
+
+<!-- 
+  To use the Avalonia CI feed to get unstable packages, move this file to the root of your solution.
+-->
+
+<configuration>
+  <packageSources>
+    <add key="AvaloniaCI" value="https://www.myget.org/F/avalonia-ci/api/v2" />
+  </packageSources>
+</configuration>
```
Kir-Antipov, line 11: Well, does it really need to be included in the template? Not everyone needs the "unstable" packages

Kir-Antipov (changes requested):
LGTM, but there're several points:

1. Replacing `$safeitemrootname$` with `$safeitemname$` will make templates a little bit more accessible.
2. Should `nuget.config` really be included in the template? I don't think that the template should in any way affect which sources users choose for their projects.

FoggyFinder:
@Kir-Antipov PR is updated

Kir-Antipov (approved):
Everything's great, I think

templates/FsAvaloniaApplicationTemplate/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("FsAvaloniaApplicationTemplate")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("FoggyFinder")]
+[assembly: AssemblyProduct("FsAvaloniaApplicationTemplate")]
+[assembly: AssemblyCopyright("Copyright © FoggyFinder 2020")]
```
rstm-sf, line 13: It seems to me that this information is superfluous :)

In addition, the new sdk does not need this file, since all [csproj can be specified](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#properties-per-attribute). It is even [possible to specify](https://www.meziantou.net/declaring-internalsvisibleto-in-the-csproj.htm) `InternalsVisibleToAttribute`. But for some reason the old sdk is also added to this pr

templates/FsAvaloniaApplicationTemplate/nuget.config
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- 
+  To use the Avalonia CI feed to get unstable packages, move this file to the root of your solution.
+-->
+
+<configuration>
+  <packageSources>
+    <add key="AvaloniaCI" value="https://www.myget.org/F/avalonia-ci/api/v2" />
```
rstm-sf, line 9: This file is actually superfluous for the new sdk. It can be replaced by `RestoreSources` from `PropertyGroup` to csproj. In addition, now, it seems to me that by default, users were using stable packages from nuget.org. Therefore, it seems that adding this file to templates is unnecessary

rstm-sf (commented):
Hello! I would like to share a few thoughts on the msbuild part

templates/FsAvaloniaApplicationTemplate/Properties/AssemblyInfo.cs
FoggyFinder: It was Is is auto-generated part. Will it be OK if I just set empty string there?

FoggyFinder: > But for some reason the old sdk is also added to this pr

That's because csproj template for templates uses old sdk. I can try to convert to the new one

FoggyFinder: I wasn't able to convert to new SDK. Reference to `Microsoft.VisualStudio.CoreUtilit` isn't resolved.

rstm-sf: Thank you!

In fact, the studio wizard can, according to what sdk will be, determine whether this file is needed or not. And so, usually in projects with a full framework, these lines still somehow change and make one single file of some kind, so these lines do not need to be added :)

rstm-sf: Oops, sorry to bother you, it's my fault. These properties were written for the template. Those. this is not for creating a project and they will not be in the created project

FoggyFinder: > These properties were written for the template. Those. this is not for creating a project and they will not be in the created project

@rstm-sf I thought you would know it :-) 
Thank you anyway - it's better to keep clean all projects

grokys (approved):
I'm no expert on F# but these templates seem to work for me! I've taken the liberty of updating the package version to the latest, we'll get this in the new version! :)

PR #177, opened 2020-10-30
Improve completion for Avalonia 0.10

donandren:
Improved completion a bit for:

- Binding path
- x:DataType
- x:CompiledBinding
- Style Setters

### DIFF
--- a/AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
@@ -93,7 +93,7 @@ namespace AvaloniaVS.IntelliSense
             {
                 if (_session == null || _session.IsDismissed)
                 {
-                    if (TriggerCompletion() && c != '<')
+                    if (TriggerCompletion() && c != '<' && c != '.' && c != ' ')
                     {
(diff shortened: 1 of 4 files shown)


AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
```diff
@@ -93,7 +93,7 @@ private bool HandleSessionStart(char c)
             {
                 if (_session == null || _session.IsDismissed)
                 {
-                    if (TriggerCompletion() && c != '<')
+                    if (TriggerCompletion() && c != '<' && c != '.' && c != ' ')
```
grokys, line 96: What case does this fix? Doesn't seem to be covered in the PR description.

grokys (approved):

AvaloniaVS/IntelliSense/XamlCompletionCommandHandler.cs
donandren: it fixed hide of Attached property hints were hidden immediately in style setter like:
```
<Style ...>
 <Setter Property="TextBlock.F
```
here without this lines the popup with hints is hidden and user can't select anything

PR #180, opened 2020-11-05
add zoom support

donandren:


AvaloniaVS/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -349,13 +371,45 @@ private async Task StartStopProcessAsync()
...
+                    Zoom = $"{Math.Round(Math.Min(x, y), 2, MidpointRounding.ToEven) * 100}%";
+                }
+                else
+                {
+                    Zoom = "100%";
+                }
+
+                return false;
+            }
+            else if (double.TryParse(Zoom.Replace("%", ""), out double zoomPercent) && zoomPercent > 0)
```
rstm-sf, line 404: Hello! Maybe using invariant mode for `TryParse` and for `ToString` by double ?

rstm-sf: And it seems cleaner to use `TrimEnd` instead of `Replace`

rstm-sf: It seems that I understood what the idea of ​​cultural dependence is: it is convenient for the user to see familiar numbers

donandren: thanks good to be little bit more on the safe side

rstm-sf: In fact, it could have been left as it is, since it happens in one culture, it seems like ... I can't find a case for changing culture. But only these lines would break everything -- it would be possible to use the same `double.ToString`

https://github.com/AvaloniaUI/AvaloniaVS/blob/d326cfb8d4988bc41c32cc317ede84da7bfb95af/AvaloniaVS/Views/AvaloniaDesigner.xaml.cs#L77-L80

AvaloniaVS/Views/AvaloniaDesigner.xaml
```diff
@@ -62,6 +62,10 @@
                       local:VsTheme.UseVsTheme="True" >
             </ComboBox>
 
+            <ComboBox Text="{Binding Zoom, ElementName=root, Mode=TwoWay}" IsEditable="True" 
+                      ItemsSource="{x:Static local:AvaloniaDesigner.Zooms}"
+                      Padding="2" local:VsTheme.UseVsTheme="True"/>
```
grokys, line 67: When I first ran this, I couldn't find the Zoom `ComboBox` as it's left-aligned while the rest of the controls are on the right. I think the following looks better:

```suggestion
            <ComboBox DockPanel.Dock="Right"
                      Text="{Binding Zoom, ElementName=root, Mode=TwoWay}" IsEditable="True" 
                      ItemsSource="{x:Static local:AvaloniaDesigner.Zooms}"
                      Padding="2" Margin="0 0 10 0" local:VsTheme.UseVsTheme="True"/>
```

image

grokys (changes requested):
Looks and works great in general! Just a few small nits in the code and a layout change suggestion.

grokys (approved):
Thanks @donandren - this is great!

PR #181, opened 2020-11-07
Add supports .NET

### LINKED ISSUE #176: does not work with TargetFramework net5.0
The extension does not seem to work when specifying `<TargetFramework>net5.0</TargetFramework>`. I get "No executable found" on the designer surface and "No Executable" / "Reference the library from an executable or wait for the solution to finish loading.". The issue seems to be in `AvalonaVS.Views.AvaloniaDesigner.LoadTargetsAsync` > `IsValidOutput` > `AvaloniaVS.Models.ProjectOutputInfo.IsNetCore`, which is implemented as follows:

```csharp
public bool IsNetCore => TargetFramework?.StartsWith("netcoreapp") ?? false;
```

rstm-sf:
Hello!

According to [this document](https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md) added .NET supports

Closes #176

### DIFF
--- a/AvaloniaVS/Services/SolutionService.cs
+++ b/AvaloniaVS/Services/SolutionService.cs
@@ -4,11 +4,12 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using AvaloniaVS.Models;
+using AvaloniaVS.Utils;
 using EnvDTE;
 using EnvDTE80;
+using Microsoft.VisualStudio.ProjectSystem;
 using Microsoft.VisualStudio.ProjectSystem.Properties;
 using Microsoft.VisualStudio.Shell;
 using VSLangProj;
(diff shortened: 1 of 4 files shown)


AvaloniaVS/Services/SolutionService.cs
```diff
@@ -210,25 +211,18 @@ private static async Task<IReadOnlyList<ProjectOutputInfo>> GetOutputInfoAsync(P
...
-                        if (!loaded.ProjectConfiguration.Dimensions.TryGetValue("TargetFramework", out var targetFw))
-                            targetFw = (await task).AllEvaluatedProperties.FirstOrDefault(p => p.Name == "TargetFramework")
-                                           ?.EvaluatedValue ?? "unknown";
+                        var hostAppNetCore = GetMsBuildProperty(msbuildProperties, "AvaloniaPreviewerNetCoreToolPath");
+                        var hostAppNetFx = GetMsBuildProperty(msbuildProperties, "AvaloniaPreviewerNetFullToolPath");
 
-                        var outputInfo = new ProjectOutputInfo
-                        {
-                            TargetAssembly = targetPath,
-                            TargetFramework = targetFw == "classic" ? "net40" : targetFw,
```
rstm-sf, line 226: Sorry, I don't know what TargetFramework this is. Didn't find it [here](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)

rstm-sf:
It's work
image

AvaloniaVS/Services/SolutionService.cs
kekekeks: It's a leftover from the old detection code that was returning "classic" for pre-netcore MSBuild style projects.

kekekeks (approved):

PR #184, opened 2020-11-28
Fixed designer crash when solution contains projects we dont like (C,…

napana:
Fixed designer crash when solution contains projects we dont like (C, C++, ...)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaVS?pullRequest=184) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaVS?pullRequest=184) before we can accept your contribution.<br/><br/>:x: napana [sign now](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaVS?pullRequest=184)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AvaloniaUI/AvaloniaVS?pullRequest=184) it.</sub>

AvaloniaVS/Services/SolutionService.cs
```diff
@@ -191,10 +192,44 @@ private static IReadOnlyList<string> GetReferences(VSProject project)
...
+#pragma warning restore IDE0059 // Unnecessary assignment of a value
+            {
+                // it seems to happen when this reference is owned by project we dont "like" (C, C++, ...)
+                //Log.Logger.Error(ex, "Error when accessing Reference.SourceProject property.");
+            }
+
+            return false;
+        }
+
+        private static bool IsNonProjectReference(Reference r)
```
FoggyFinder, line 216: it is the same as 

```cs
private static bool IsNonProjectReference(Reference r) => !IsProjectReference(r);
```

Are you sure it's a good idea to add such function? Especially when there is only 1 usage.

napana: If exception happens in IsProjectReference than it returns false and that doesn't mean the reference is non project one so i created two methods to avoid this negation.

FoggyFinder: Oh, that's right. Didn't look carefully.

napana:
Need to do this again because it now contains my other commit - doing it second time in my life sorry :)

PR #187, opened 2020-11-29
Added Avalonia Class Library project template

napana:
This PR depends on #185

This PR introduces new project template Avalonia Class Library.


Takoooooo:
Sorry, I don't think we need it. Also our AvaloniaVS templates are fully compatible with our CLI templates so every new template should be added to CLI templates first

Issue #190, opened 2020-12-19
Designer takes the assembly only from Debug


rstm-sf:
Hello!

Designer takes the assembly only from Debug. If I choose Release, it still takes from the Debug

MikeCodesDotNET:
I've had a quick look at this and think I've found a less than ideal workaround. 

If you change the build configuration to Release, unload the project, and then reload, a call to get the [TargetPath](https://github.com/AvaloniaUI/AvaloniaVS/blob/fcb79cdd172effbd9efb790ff1281d6f38c1cb19/AvaloniaVS/Services/SolutionService.cs#L214) will occur. This will then ensure the design is loading from the Release directory. 

I'll have a look to see how we can detect build configuration changes and then do this automagically. If I can get it resolved, then I'll submit a PR.

maxkatz6:
This bug was fixed in recent VS extension builds. Please let us know if you still have issues.

PR #196, opened 2021-02-15
Fix for #190 - taking assembly from debug rather than active configur…

MikeCodesDotNET:
This is my first attempt at fixing #190

**Do not merge. Can we discuss?**

I'm not expecting it to be merged (please don't). Instead, I'm looking for an opinion / guidance that I'm heading in the right direction with this fix.

I'm thinking that if hooking into IVsUpdateSolutionEvents is correct, then perhaps SolutionService should implement it so I can remove UpdateSolutionEvents class entirely. 

I'm still not 100% sure why anyone would need the designer assembly to come from the active configuration, but it's an interesting excuse for me to dig into the extension anyway.

### DIFF
--- a/AvaloniaVS/AvaloniaPackage.cs
+++ b/AvaloniaVS/AvaloniaPackage.cs
@@ -60,6 +60,7 @@ namespace AvaloniaVS
         public const string Name = "Avalonia Xaml Editor";
 
         public static SolutionService SolutionService { get; private set; }
+        public static UpdateSolutionEvents UpdateSolutionEvents { get; private set; }
 
         protected override async Task InitializeAsync(
             CancellationToken cancellationToken,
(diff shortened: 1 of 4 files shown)


grokys:
Apologies taking so long getting back to you on this: I'm not sure any of us are particularly expert on the VS extension, we just hack things together because we need to ;)

I think `SolutionService` implementing `IVsUpdateSolutionEvents` _probably_ sounds correct?

> I'm still not 100% sure why anyone would need the designer assembly to come from the active configuration, but it's an interesting excuse for me to dig into the extension anyway.

Me neither, this is #190, right? We probably need more info as to why they want the designer to use the Release config. Using always Debug feels intuitively more "right" to me but I don't feel particularly strongly about that.

MikeCodesDotNET:
> not sure any of us are particularly expert on the VS extension, we just hack

I'm fairly sure this is the case for 99% of people trying to build VS extensions! It's amazing how complex it all seems to be. 

Perhaps @rstm-sf can enlighten us as to the reason behind needing the designer to use the release builds?  I'm inclined not to update the implementation in preparation for review until we can agree on if this is really needed (I suspect it might not be).

rstm-sf:
Hi, sorry for the delay.

Indeed, for 99% it will be really unnecessary. For example, I accidentally discovered this when I decided to change to Release, then decided to change the xaml and see what changed, but nothing happened... I agree that this is an corner case

For example, as far as I remember, in the [AvaloniaRider](https://github.com/ForNeVeR/AvaloniaRider), when every time users select a xaml-file, settings are taken from msbuild

PR #216, opened 2021-10-21
feat: Allow to discover the internal classes and it is InternalsVisibleTo

### LINKED ISSUE #130: Autocomplete shouldn't list internal properties
For example. `RowDefinitionsDirty`:

sample

If one have set the property then runtime exception will be thrown:

```
Unhandled exception. System.MethodAccessException: Attempt by method 'Views.MainWindow.!XamlIlPopulate(System.IServiceProvider, Views.MainWindow)' to access method 'Avalonia.Controls.Grid.set_RowDefinitionsDirty(Boolean)' failed.
```

### LINKED ISSUE #215: Autocomplete should list internal class
**To Reproduce**

1. Create class like this:
    ```csharp
      using Avalonia;
      
      namespace Grid_IsVisible_Issue.Behaviors
      {
          static class ColumnDefinition
          {
              private readonly static Avalonia.Controls.GridLength ZeroWidth = 
                 new Avalonia.Controls.GridLength(0,Avalonia.Controls.GridUnitType.Pixel);
      
              private readonly static AttachedProperty<Avalonia.Controls.GridLength?> LastWidthProperty =
                  AvaloniaProperty.RegisterAttached<Avalonia.Controls.ColumnDefinition, Avalonia.Controls.GridLength?>("LastWidth"
                      , typeof(ColumnDefinition)
                      , default );
      
              public readonly static AttachedProperty<bool> IsVisibleProperty =
                   AvaloniaProperty.RegisterAttached<Avalonia.Controls.ColumnDefinition, bool>("IsVisible"
                       , typeof(ColumnDefinition)
                       , true
                       , coerce: ((element, visibility) =>
                            {
      
                                var lastWidth = element.GetValue(ColumnDefinition.LastWidthProperty);
                                if (visibility == true && lastWidth is { })
                                {
                                    element.SetValue(Avalonia.Controls.ColumnDefinition.WidthProperty, lastWidth);
                                }
                                else if (visibility == false)
                                {
                                    element.SetValue(LastWidthProperty, element.GetValue(Avalonia.Controls.ColumnDefinition.WidthProperty));
                                    element.SetValue(Avalonia.Controls.ColumnDefinition.WidthProperty, ZeroWidth);
                                }
                                return visibility;
                            })
                       );
      
              public static bool GetIsVisible(Avalonia.Controls.ColumnDefinition columnDefinition)
              {
                  return columnDefinition.GetValue(IsVisibleProperty);
              }
      
              public static void SetIsVisible(Avalonia.Controls.ColumnDefinition columnDefinition, bool visibility)
              {
                   columnDefinition.SetValue(IsVisibleProperty,visibility);
              }
          }
      }
      ```
2.  Add xml namespece
    ```xaml
      <Window xmlns="https://github.com/avaloniaui"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:vm="using:Grid_IsVisible_Issue.ViewModels"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
              x:Class="Grid_IsVisible_Issue.Views.MainWindow"
              Icon="/Assets/avalonia-logo.ico"
              Title="Grid_IsVisible_Issue"        
              xmlns:lb="using:Grid_IsVisible_Issue.Behaviors">
      
    ```
3. Try add to ColumnDefinition
    immagine

if made class public is show

immagine

### LINKED ISSUE #235: Better filtering for Intellisense completion suggestions
Currently Intellisense suggestions often suggest things that are completely irrelevant or are "internal" to Avalonia. Biggest offender is typing ContentTemplate, and ContentTemplateResult is the first suggestion. HorizontalContentAlignment and VerticalContentAlignment are always prioritized over HorizontalAlignment and VerticalAlignment - IMO the shorter name properties should be displayed first. These are just a couple that immediately come to mind, there are others though. In both these cases, I think sorting the completion list would solve at least the immediate issue - even if not a great fix overall.

image

Typing a markup extension like {StaticResource } always produces a list of Default/Simple theme resources. It'd be nice if this could not do this, particularly if the Default/Simple theme isn't used. Even better if it could read the current resources available in the current App, or if loading from the entire solution may be a perf problem, at least the current document and App.axaml should be manageable. 

image

I wanted to try to tackle this myself, but I can't wrap my head around how intellisense works, particularly because its seems it works both here in the extension and the Avalonia IDE submodule, and it's not clear how they link together.

workgroupengineering:
fixes #215 
fixes #130 
fixes #235

part of # #279


workgroupengineering:
Witing https://github.com/kekekeks/Avalonia.Ide/pull/38 approved

CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs
```diff
@@ -652,36 +672,36 @@ IEnumerable<string> filterLocalRes(MetadataType type, string currentAssemblyName
 
             bool isbitmaptype(string resource) => bitmaptypes.Any(ext => rhasext(resource, ext));
 
-            if (types.TryGetValue("Avalonia.Media.Imaging.IBitmap", out MetadataType ibitmapType))
+            if (types.TryGetValue(key => key.StartsWith("Avalonia.Media.Imaging.IBitmap"), out MetadataType ibitmapType))
```
maxkatz6, line 675: I checked why did you need to replace dictionary lookup with loop iterations in many places. And that's because key now is a type's fully qualified name.
But typical simple Avalonia.Samples app has about 2500 types, and there is multiple `TryGetValue(StartsWith)` calls.
Can it be optimized with second dictionary? I think you could do it in PostProcessTypes's scope by iterating over all of these types and creating another local dictionary once.

tests/CompletionEngineTests/Metadata/MetadataConverterTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            MetadataType typeA2 = ns[clrTypeA2.Name];
+
+            Assert.NotNull(typeA2);
+
+            Assert.Equal(clrTypeA2.AssemblyQualifiedName, typeA2.AssemblyQualifiedName);
+        }
+
+
+        [Fact]
+        public void Discover_InternalsVisibleTo()
```
maxkatz6, line 73: Internal members from other assemblies are still visible, when should not be (https://github.com/AvaloniaUI/AvaloniaVS/issues/130):
image

Other issue with internal types in the same assembly seems to be fixed.

maxkatz6 (changes requested):
I am not an expert with this area of code but attempted to review it.
Found some problems.

maxkatz6 (approved):

Issue #206, opened 2021-06-20
Support Visual Studio 2022


MikeCodesDotNET:
As per the title. We need to add support for VS22. 

I've already got a local branch with it working and it was pretty straight forward to upgrade. No code changes were required!

MikeCodesDotNET:
I've commited https://github.com/AvaloniaUI/AvaloniaVS/commit/d2e3888e6dedf3d5fa234c56044ac56afebc4043 which enables support for the 17.0.0 Preview 1.0 [31408.342.d17.0]. I believe my version in an internal build, so the changes will need to be checked on the public preview release.

maxkatz6:
@MikeCodesDotNET it seems "[17.0.0-previews-1-31410-273](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK/17.0.0-previews-1-31410-273)" is in public preview

0x90d:
Any update?

stesmi:
I cloned the repo and updated references and I have something running. I have not tested it extensively yet, but right now I have ReSharper and Avalonia running in VS 2022 preview 3.

My changes were way bigger than Mike's, though, and I haven't checked how many of the updates are actually necessary.
Hopefully we can get an official release soon.

grokys:
I've been looking into this for VS 2022 Preview 3.1 and our use of the DTE seems to be causing problems, looks like we need to replace our usage of that with something more modern.

I've not spent much time on it so far but I've opened a branch at https://github.com/AvaloniaUI/AvaloniaVS/tree/update/vs-2022 with my progress so far.

As an aside, should we drop support for < VS2019? I don't have VS2017 installed on my machine any more to test with.

uberhalit:
As final VS2022 released now, I'd be great if we can get an official fox soon.

jonfast565:
Really interested in seeing a fix for this soon!

grokys:
@uberhalit @jonfast565 we have an installer which works in #223, and you can already download an installer from our CI (https://dev.azure.com/AvaloniaUI/AvaloniaUI/_build/results?buildId=16490&view=artifacts&pathAsName=false&type=publishedArtifacts).

The remaining problem is to get this working on earlier versions too.

Related: https://github.com/AvaloniaUI/Avalonia/issues/6927

alexhelms:
[...]

PR #223, opened 2021-11-09
VS2022 Support

### LINKED ISSUE #222: Does not install on Visual Studio 2022
Searching for the extension in the IDE doesn't return it, attempting to install it by downloading a .vsix throws the `VSIXInstaller.NoApplicableSKUsException` exception. Here's the installer log: [dd_VSIXInstaller_20211105135926_22cc.log](https://github.com/AvaloniaUI/AvaloniaVS/files/7486143/dd_VSIXInstaller_20211105135926_22cc.log)

Takoooooo:
If you are reading that and have VS2022/VS2019/VS2017 installed please test the accompanying DevOps build artefact, it will speed up merging this PR.
closes #222,#206


Stephanvs:
I've downloaded the accompanying DevOps build artifact, and installed the VS extension successfully in both:
```
Microsoft Visual Studio Professional 2022 (64-bit) - Current
Version 17.0.0```

and:

```
Microsoft Visual Studio Professional 2022 (64-bit) - Preview
Version 17.1.0 Preview 1.0
```

And I've verified that the extension works by opening up an .axaml file, and the previewer seems to work.

Hope this helps
Stephanvs

grokys:
Gave this a quick try on VS2022 Preview 1.0 and seems to work well. However, I'm not given the option to install it on VS2019 - this should still be supported, right?

Takoooooo:
> Gave this a quick try on VS2022 Preview 1.0 and seems to work well. However, I'm not given the option to install it on VS2019 - this should still be supported, right?

Yes,and even VS2017 should be supported

KieranDevvs:
> > Gave this a quick try on VS2022 Preview 1.0 and seems to work well. However, I'm not given the option to install it on VS2019 - this should still be supported, right?
> 
> Yes,and even VS2017 should be supported

Id assume this is due to the prerequisite targeting .NET 6 and that not being supported in VS2019 and lower?
Maybe we have to have two extensions?

grokys:
@KieranDevvs where is there a prerequisite targeting NET6? That would indeed cause this problem I guess, only problem is I can't see it!

Gillibald:
Looks like someone found a solution to this: https://cezarypiatek.github.io/post/migrate-vsix-to-vs2022/

HavenDV:
If anyone wants to test the extension in VS2022, here is the file:
image
[AvaloniaVS.zip](https://github.com/AvaloniaUI/AvaloniaVS/files/7681249/AvaloniaVS.zip)

KieranDevvs:
Why has this been closed?

PR #227, opened 2021-11-24
Fix "When you open a axaml\xaml file in VS with this extension it will appear in VS git as changed "

### LINKED ISSUE #220: When you open a axaml\xaml file in VS with this extension it will appear in VS git as changed
It is really annoying.
https://recordit.co/fhSqN5BMUb
When you close the tab it will be removed from the list of the changes.

Takoooooo:
This code fixes the problem on vs2019, on vs2022 this problem doesn't appear so that should be a VS bug. Btw I am not sure whether this will break something or not because I am not familiar with this code.
closes #220

### DIFF
--- a/AvaloniaVS/Services/EditorFactory.cs
+++ b/AvaloniaVS/Services/EditorFactory.cs
@@ -141,7 +141,7 @@ namespace AvaloniaVS.Services
 
                 var guidIVSTextLines = typeof(IVsTextLines).GUID;
                 ErrorHandler.ThrowOnFailure(invisibleEditor.GetDocData(
-                    fEnsureWritable: 1,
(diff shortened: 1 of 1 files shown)


grokys:
Not sure either, the docs say https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsinvisibleeditor.getdocdata?view=visualstudiosdk-2017

> If the caller intends to modify the document, they should pass TRUE for fEnsureWritable, which queries the user for checking the file out as appropriate if it is under source code control.

This sounds kinda outdated and related to VSS or something. I guess we just need to try it and see if it causes problems?

grokys (approved):
Note sure either if it will break anything

PR #229, opened 2021-12-13
VS2022 Support

### LINKED ISSUE #206: Support Visual Studio 2022

Takoooooo:
This pull request adds support for VS2022 and fixes all issues we had with VS2022 integration. I moved all contents to the Shared project and added two projects, one for VS2022, one for VS2019\VS2017 and fixed pipeline to add both artefacts.
closes #206


grokys (approved):
Seems to work on all the versions of VS I have available!

PR #230, opened 2021-12-15
VS2022 Extension Tweaks

grokys:
- We need to update publisher to match the publisher on marketplace
- VS2022 extension needs a separate ID - use `AvaloniaVS` as it's cleaner than `AvaloniaForVisualStudio2022`
- Removed directory that should have been deleted in #229


Takoooooo (approved):
LGTM,thanks!

PR #237, opened 2022-02-11
Fix #231 - Ctrl-F not opening the quick search box in VS2022

### LINKED ISSUE #231: Under VS2022, in a styles file, ctrl-f opens a search window, and not the file's search box at the top right corner
ctrl-f should open a search box at the top right for the file not the search dialog for the whole solution. 

Seems like ctrl-f works fine in a XAML resource files, only Styles files have problems. 

This problem is unique to VS2022 - everything works fine under VS2019.

odalet:
This should fix #231 

As explained [here](https://github.com/AvaloniaUI/AvaloniaVS/issues/231#issuecomment-1032380289), I'm not very proud of this, but hey... it seems to work.

I've ensured it fixes the issue in VS2022 and it keeps working in VS2019 (could not test in VS2017, but there's no reason it shouldn't work)

PS: I also took the liberty to remove the reference to `AvaloniaVS.VS2019AndEarlier` from the `AvaloniaVS.VS2022` as it only had Visual Studio scream about duplicate symbols... If there was a legit reason for this reference, I'll add it again of course.

### DIFF
--- a/AvaloniaVS.Shared/Services/IVsFindTarget3.cs
+++ b/AvaloniaVS.Shared/Services/IVsFindTarget3.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.Runtime.InteropServices;
 
-namespace AvaloniaVS.Services
+// Do not change the namespace; otherwise, search box is broken in VS2022
+namespace Microsoft.VisualStudio.Editor.Internal
 {
     [ComImport]
+    [TypeIdentifier]
     [Guid("A2F0D62B-D0DD-4C59-AAB8-79CD20785451")]
     [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
     public interface IVsFindTarget3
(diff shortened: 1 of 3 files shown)


Takoooooo:
Thanks for looking at this issue, you probably understand now why we don't do almost anything on VS extension... I will try to review your work ASAP

odalet:
@Takoooooo well yes, VS extensions are a bit of a mess, I'm no expert at all in those, but, and I won't promise anything, I may try to experiment with refactoring the code to have something similar to the markdown extension I talked about. The author (Mad Kristensen) is well known for the quality of his extensions. 

Btw, would it be useful to update the sub-modules to a more recent commit of Avalonia? It seems to work just fine now, but I'm still wondering.

Takoooooo:
@odalet We would be very happy if someone will help us improving VS extension, we don't have many resources currently to maintain it, but it definitely needs improvements. If you will help us with that I will try to find people who will review your code\review it by myself.

Takoooooo:
Thanks for the contribution. We don't have any idea how that works but decided to merge that since it works

PR #239, opened 2022-02-17
Previewer Improvements

### LINKED ISSUE #236: Previewer improvements
It would be nice if there was some padding around the Image element that hosts the preview Image. Basically making it closer to WPF where you can essentially scroll the image off screen. If an element is on the edge of the control/window preview, you have no way of centering it in the preview pane because the ScrollViewer is locked to the image size. I think adding something like Padding = 500 (or whatever) to the ScrollViewer and centering the ScrollViewer when it loads would solve this easily.

The preview host flashes anytime it is updated - which can be quite distracting if you make multiple edits quickly. It be a much better experience if the Image was kept as it's being replaced. 

As a side note: Wanted to look at both of these myself, but I'm getting AvaloniaPackage failed to load anytime I open up an axaml file while debugging the extension. git submodule update --init --recursive and dotnet restore successfully and no errors, but I just get the normal xml editor and using open with gives a popup with that error message.

amwx:
A couple improvements to the VS previewer to make it a little closer to the WPF/UWP designer experience:
- Added ability to scroll beyond the preview image. It's not the same as the WPF/UWP designer since it isn't infinite scroll, but just adding a margin to the previewer image to make it so you can scroll beyond it (helpful if you you want to center the edge of the previewer while working on something). Made sure changing zoom, design size, etc. still provides a decent experience by re-centering the ScrollViewer after a size change. 
- Added ability to scroll horizontally with Shift+MouseWheel. Wanted to also add zoom with Ctrl+MouseWheel, but there's a really bad lag sometimes with changing zoom and it wasn't a great experience.
- Removed the white background behind the preview image. This helps with the flashing when making text edits with VS in dark mode, but still doesn't solve that issue. That problem seems to be coming from upstream with two bitmaps being sent, one being completely transparent. Will open a related issue on Avalonia repo about that since I think that's where the solution is.
- Hides the previewer during debugging (now only Xaml shows), consistent with WPF/UWP
- Fixed the error states in AvaloniaDesigner, particularly the invalid markup error text to show the Invalid Markup error - which wasn't running because it checked for `Process.Bitmap == null && Process.Error != null`, which seems to never be the case - so changed the `&&` to `||`. Also in FrameReceived, checked that Error was null in addition the bitmap check

Fixes #236 

Tested with VS2022, I haven't tested with VS2019, but I don't think anything here would break that extension version


Takoooooo:
Thx for the contribution,i will try to review it ASAP!

Takoooooo:
Btw,i think it should fix #236 instead of #239?

Takoooooo:
Maybe you have a telegram or just any way where I can contact you directly? I would really like to help you and finish that PR :)
My telegram is @takooooooooo

Takoooooo:
btw, i am not sure how to reproduce the problems that were fixed by these fixes,may you point me?
image

amwx:
> Can we make the horizontal scroll a bit faster? Look at this video, it scrolls vertically much faster than horizontally.Also, can you add scrolling back and further with ctrl+mousewheel as in WPF? Btw I can help you if you want :) https://recordit.co/4kH6KlujiI Edit: Oh I see you have tried adding ctrl+mousewheel scrolling, maybe you have the code which you have tried so I may take a look?

Yes I will look look at that - actually meant to, but forgot


> btw, i am not sure how to reproduce the problems that were fixed by these fixes,may you point me? 

The first, the flashing issue can be seen in this gif (apologies for the low res): The white flash is because there was a white background applied to the Border that hosts the Image control. The whole flashing thing, as I've noted, is because two FrameReceived events are being sent for every Xaml update - the first being transparent, the second being the actual updated image. I verified this by writing the bitmaps to file. But I don't know why because this traces back to one of the submodules, which VS refuses to load symbols for, and its all asynchronous making it harder to trace how updates happen

FluentAvalonia - Microsoft Visual Studio 2022-02-18 17-30-22


The second issue, is just something I noticed. I realized that I don't recall seeing the "Markup Error" error text, instead usually the previewer just disappears if you have an error:

image

After putting a breakpoint in, I noticed that the previewer bitmap is never null when there's an error - so that condition was always failing.


> Btw,i think it should fix #236 instead of #239?
Oops. That would explain why the PR wasn't linking to my issue 😂

Takoooooo:
Haven't investigated that deeply but seems that extension isn't working with this PR on VS2019(

amwx:
@Takoooooo I've fixed horizontal scrolling - made it a factor of the mouse wheel delta and it seems pretty close to vertical now. Also tested with VS2019 and all seems to work on my end.

Takoooooo:
Thanks a lot for this contribution! It really improved Avalonia designer!

PR #246, opened 2022-03-21
Add ability to zoom in and out with ctrl + mousewheel

Takoooooo:
It works in the same way as WPF previewer does.

There is a problem that sometimes it scrolls with a small lag because all scrolling logic is async and it is using FireAndForget, but anyways IMO it is better than nothing.


grokys (approved):

SuperJMN:
Wonderful!

PR #267, opened 2022-09-28
Remove templates from extension

Takoooooo:
When [this](https://github.com/AvaloniaUI/avalonia-dotnet-templates/issues/108) feature will be added we can safely remove templates from the extension and just make people use the CLI templates instead.


Mrxx99:
But it would be a bad experience if by installing the VS extension, you would not get any templates.
Also besides than having to install the templates separately I think you still would need to go to Visual Studio settings first to enable that CLI templates are shown in VS (or is this now the default?)

Takoooooo:
> But it would be a bad experience if by installing the VS extension, you would not get any templates. Also besides than having to install the templates separately I think you still would need to go to Visual Studio settings first to enable that CLI templates are shown in VS (or is this now the default?)

In VS2022 CLI templates are shown by the default,in VS2019 you should enable that additionally but i believe that the majority of community are already using VS2022. I think having templates separetly makes sense because we don't need to maintain two repos with the templates at the time and i am pretty sure we may in-built CLI templates installation into VS extension

Takoooooo:
Closing because this PR is outdated, ItemTemplates need to be included and there are conflicts with the master branch already

Mrxx99:
@Takoooooo Reagrding item templates you can upvote this issue and maybe share it with Avalonia team to get some upvotes: https://developercommunity.visualstudio.com/t/dotnet-CLI-item-templates-not-supported/10091295?q=item+template&page=1

PR #272, opened 2022-12-03
Add VS specific CompletionEngine

### LINKED ISSUE #269: Merge Avalonia.IDE with the main extension repo
Avalonia.IDE hasn't seen a commit in 2 years now and the open PRs haven't been responded to which makes it feel its not actively maintained anymore. Which is fine if that's the case (because of time or whatever), but the Xaml experience in VS for Avalonia is starting to degrade. There's several issues already known (some with pending fixes) and with the new ControlTheme stuff, intellisense doesn't properly initialize within ResourceDictionaries making writing control themes a bit more cumbersome. 

I think it'd be best to remove the submodule for Avalonia.IDE and merge the completion stuff with this repo so it's all together. This has two advantanges. First, it makes adding fixes easier for community members like myself since we don't need to do fixes in multiple places hoping they get merged upstream and then here. Second, it ensures the entire extension is actively maintained and can get fixes or improvements as needed.

amwx:
This PR adds a VisualStudio extension specific version of the CompletionEngine from Avalonia.IDE.CompletionEngine. This is needed to help advance the intellisense experience with this extension. I didn't remove the submodule to Avalonia.IDE.Completion and tried to preserve what could be, but some types have been copied over that are needed. This may result in duplicate code or the code diverging if upstream is ever updated, but I think that's a fair tradeoff for potentially faster development of completion within VS.

I've also, as a demo, added completion for pseudoclasses within Style Selectors:
ezgif com-gif-maker(1)

This is done by reading the `PseudoclassesAttribute` on supporting types. It's done through a nasty way, but it works - would be better if we had access to the `PseudoclassesAttribute` type to directly query that. I've also adjusted some of the completion logic so that pseudoclasses trigger a new completion session so typing "Control:" will list the pseudoclasses, previously because there's no space after 'Control', the engine wouldn't know to refilter the hints and we wouldn't get a list of suggestions since nothing matches "Control:". I've also made the char `-` NOT end a completion session for pseudoclasses like `:focus-within` allowing that to be typed without interruption. 

I've also fixed handling completion commit on '.' key. Currently typing 'Window.Resources', the '.' will cancel the completion session and swallow the '.' character so you'd have to press it twice to continue typing, which is suboptimal. Now the '.' character isn't swallowed and a new completion session is activated after pressing it.

Edit: I've also fixed the double intellisense popup when adding xml namespaces. I think VS XML service was starting its own and then we were starting one ourselves which was causing the double popups. The existing completion session already had an Avalonia completion set in it, so instead of creating a new session, I've made it just tap into that one.
image

TODO:
- [x] Fix intellisense suggestions being suboptimal in ControlThemes
  - As an example, in ControlTheme, typing `<Setter Property="Tem` will suggested `TemplatedParent` and `TemplatedControl` but not the property `Template` (which is what's desired)

Future TODOs (not in this PR):
- find and list all defined resources instead of using static, hard-coded ones (hopefully, if not will add to future todos) that don't really apply to anything outside the old default theme
  - Ideally with this we need to make sure at least the active document (maybe even the current project) can update the suggestions more frequently without requiring a build (which is how the metadata is currently invalidated) as if you add a resource to the current file, I would expect it to immediately show without having to build the project. External assemblies could probably be cached like normal on build only since they're static unless the ref is explicitly updated

- Filter out non-public types from suggestions
- Add better context awareness for suggestions, including:
  - filtering the suggestions to only include what makes sense at the given point in the document (this is mostly already how it is, but I think there are a few places where this could be improved, as an example, don't suggest `VerticalContentAlignment` unless it absolutely makes sense to do so (only suggest `VerticalAlignment`)
  - In a Style Selector, we could probably make it only suggest things that make sense given the cursor position, ie don't suggest `/template/` or pseudoclasses at the start of the selector, no pseudoclasses directly after `/template/`, etc
  - In this PR, I've also added some code to tell VS to auto select the first suggestion in the list - previously, it wouldn't usually do this and you had to physically interact (mouse or arrow keys) to get the completion to work. We could probably improve this to guess the best item and select it instead based on context
- Fix editing existing text and accepting a completion
  - currently if you type `ponterover` and try to correct it, adding the `i` will suggest ":pointerover", but accepting that completion will end up with `pointerovernterover` as it just inserts the suggested text and doesn't check anything for replacing

closes #269

### DIFF
--- a/AvaloniaVS.Shared/Completion/AvVSCompletionEngine.cs
+++ b/AvaloniaVS.Shared/Completion/AvVSCompletionEngine.cs
@@ -1,0 +1,662 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Avalonia.Ide.CompletionEngine;
+using System.Text.RegularExpressions;
+using System.Xml;
+using AvMetadata = Avalonia.Ide.CompletionEngine.Metadata;
+using AvaloniaVS.Shared.Completion.Metadata;
+using AvCompletion = Avalonia.Ide.CompletionEngine.Completion;
+
+namespace AvaloniaVS.Shared.Completion
+{
+    public class AvVSCompletionEngine
+    {
+        private class MetadataHelper
+        {
+            private AvMetadata _metadata;
+            public AvMetadata Metadata => _metadata;
+            public Dictionary<string, string> Aliases { get; private set; }
+
+            private Dictionary<string, MetadataType> _types;
+            private string _currentAssemblyName;
+
+            public void SetMetadata(AvMetadata metadata, string xml, string currentAssemblyName = null)
+            {
+                var aliases = GetNamespaceAliases(xml);
+
+                //Check if metadata and aliases can be reused
+                if (_metadata == metadata && Aliases != null && _types != null && currentAssemblyName == _currentAssemblyName)
+                {
+                    if (aliases.Count == Aliases.Count)
+                    {
+                        bool mismatch = false;
+                        foreach (var alias in aliases)
+                        {
+                            if (!Aliases.ContainsKey(alias.Key) || Aliases[alias.Key] != alias.Value)
+                            {
+                                mismatch = true;
+                                break;
+                            }
+                        }
+
+                        if (!mismatch)
+                            return;
+                    }
+                }
+                Aliases = aliases;
+                _metadata = metadata;
+                _types = null;
+                _currentAssemblyName = currentAssemblyName;
+
+                var types = new Dictionary<string, MetadataType>();
+                foreach (var alias in Aliases.Concat(new[] { new KeyValuePair<string, string>("", "") }))
+                {
+                    Dictionary<string, MetadataType> ns;
+
+                    string aliasValue = alias.Value ?? "";
+
+                    if (!string.IsNullOrEmpty(_currentAssemblyName) && aliasValue.StartsWith("clr-namespace:") && !aliasValue.Contains(";assembly="))
+                        aliasValue = $"{aliasValue};assembly={_currentAssemblyName}";
+
+                    if (!metadata.Namespaces.TryGetValue(aliasValue, out ns))
+                        continue;
+
+                    var prefix = alias.Key.Length == 0 ? "" : (alias.Key + ":");
+                    foreach (var type in ns.Values)
+                        types[prefix + type.Name] = type;
+                }
+
+                _types = types;
+            }
+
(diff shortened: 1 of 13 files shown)


amwx:
This is pretty much ready at this point. There's a lot more I'd like to do (as described above), but I think that's better delegated to separate PRs. 

Listing resources from the assemblies is probably top of the list, but I don't really know how to do this since I'm not familiar with how to get the AvaloniaResources without the AvaloniaXamlLoader which is both not available, and overkill. Any suggestions or help is welcome.

grokys:
Other than that looks good to me I think.

amwx:
> @kekekeks are you happy to fork the completion engine into this repository for now? I assume you don't have much time to maintain it anyway.

@grokys FYI I already got approval in the original issue ([this comment](https://github.com/AvaloniaUI/AvaloniaVS/issues/269#issuecomment-1298268923)). Still probably wouldn't hurt to still get a second approval though

AvaloniaVS.Shared/Completion/AvVSCompletionEngine.cs
```diff
@@ -0,0 +1,678 @@
...
+
+                        completions.AddRange(GetHintCompletions(prop.Type, search, currentAssemblyName));
+                    }
+                    else if (prop?.Type?.Name == typeof(Type).FullName)
+                    {
+                        completions.AddRange(_helper.FilterTypeNames(state.AttributeValue).Select(x => new AvCompletion(x, x, x, CompletionKind.Class)));
+                    }
+                    else if (state.AttributeName == "xmlns" || state.AttributeName.Contains("xmlns:"))
+                    {
+                        if (state.AttributeValue.StartsWith("clr-namespace:"))
```
maxkatz6, line 330: Note this PR doesn't have changes from here https://github.com/kekekeks/Avalonia.Ide/pull/36/files
Which are actually used in this PR (you can see from the submodule).

amwx: Good catch. The submodule is obviously ahead of master then, which is where I copied everything from.

kekekeks:
@amwx IIRC the idea was to keep the engine in a separate library that can be used outside of visual studio, even if it' in the same repo

grokys:
Yeah putting the library into AvaloniaVS makes sense I think, but if we could keep it as a separate assembly that has no dependencies on VS that would be good, as we may want to extract it in future to use for other IDEs.

amwx:
This is actually what I wanted to do, but I clearly misunderstood @kekekeks's comment in the issue then. I'll revert this and move the completion engine as a whole here - keeping it as a separate project/assembly for external use. Will convert back to a draft for the time being.

FWIW, I don't really care where the completion engine is housed, but my concern was the lack of attention it was getting and intellisense/completion getting worse in VS. So as long as its somewhere it's likely somebody can review PRs and get fixes merged I'm fine with it, as intellisense/completion are critical parts of IDEs. As VS user though, housing it here makes it easier to test.

PR #274, opened 2022-12-06
EditorFactory fix: properly handle uninitialized IVsTextBuffers when initializing the editor

amwx:
Currently, there's an issue with this extension that if you leave multiple axaml files open when closing visual studio, only one of them will actively reopen. The others you have to manually close and reopen. In addition, the one tab that does reopen is reordered in the tab bar because VS was told to close & reopen the document internally. And we're left with this sad screen:
image

This has been an issue for a while and no one has been able to figure out why or the solution. This has been one of my biggest issues with this extension (and to no fault at all to the Avalonia team), so has been a side-project of mine for many months (off and on) of lots of trial and error. And I think I've finally found it.

Currently, we're using the invisible editor approach to initialize an IVsTextBuffer object for the xaml file we want to open. That's fine, except this error is originating in the call to `IVsInvisibleEditor.GetDocData` which is throwing an `E_INVALIDARG` HRESULT. And because this is COM/C++, closed-source land for the core of VS, it falls into the category of "unhelpful error message". The issue where the one file reopens but the tab is reordered was caused by returning `VS_E_INCOMPATIBLEDOCDATA` because the call to `eafs.GetDocData()` was failing on an uninitialized `IVsTextLines`. The root cause of all of this is some internal change that IVsTextBuffers aren't initialized right away. And the invisible editor provides no way to get the buffer until its initialized so the question becomes, "well what now?"

From what I can tell, the invisible editor approach has been silently "deprecated" in favor of creating your own IVsTextBuffer, which it turns out is quite easy to do and provides us a way to wait for initialization. These two samples show how to do this and were used:
https://github.com/madskristensen/EditorConfigLanguage/blob/master/src/LanguageService/EditorFactory.cs
https://github.com/microsoft/VSSDK-Extensibility-Samples/tree/master/WPFDesigner_XML/WPFDesigner_XML

After setting up the code to wait for initialization, a new error arose in that VS doesn't like to initialize a IVsCodeWindow with an uninitialized text buffer and it causes the designer to entirely abort loading our previewer and display some unhelpful error message linking lines to the internal workings of VS (genius right?). The fix for this is to delay creating the IVsCodeWindow until the text buffer is fully created. This required reworking much of the editor pane stuff to account for this, and delaying the start of the previewer process. I've tried to add as many comments as possible to document what I did for future reference, but if anything needs clarification, I'll do my best.

The end result of all of this, however, is that you can now leave as many xaml files open as you want and they will all restore properly - and none of them are reordered because we never return `E_INCOMPATIBLEDOCDATA` triggering VS to reload the document. 

As far as I can tell this is working quite well, but definitely needs to be tested by others. I don't have VS2019 any more, so I can't test that, but considering this issue started in VS2019 and all the SDK functions are available this should work just fine there too.

<s>Known regression:
- The context menu will no longer show the keyboard shortcuts for "View Code" and "View Designer"</s>- though these function didn't really work anyway 

<s>NOTE: I'm only leaving WIP in the title to see if I can address the regression & make show code/show designer work. Otherwise, the PR is ready.</s>
EDIT: It seems its the experimental/test version of VS is responsible for not showing the keyboard shortcuts. However, I did fix "View Code" so that it properly opens the code file for the xaml file. Currently it does require that VS is properly nesting the code file under the xaml file though.


AvaloniaVS.Shared/Commands/ViewCodeCommandHandler.cs
```diff
@@ -15,24 +16,39 @@ namespace AvaloniaVS.Shared.Commands
...
         public bool ExecuteCommand(ViewCodeCommandArgs args, CommandExecutionContext executionContext)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
             var dte = (DTE)AvaloniaPackage.GetGlobalService(typeof(DTE));
 
+            var activeDocument = dte.ActiveDocument;
+
+            // We only want to handle "View Code" if our designer pane is active
+            if (!activeDocument.FullName.EndsWith(".axaml", StringComparison.OrdinalIgnoreCase))
```
maxkatz6, line 34: what about ".xaml"?

amwx: Fixed

maxkatz6 (approved):
Thanks!
Tested with 2022 and it works well for me. Can't check 2019 either.

grokys:
Thanks for fixing this @amwx !

PR #275, opened 2022-12-08
Move Avalonia.Ide.CompletionEngine into this repo

### LINKED ISSUE #269: Merge Avalonia.IDE with the main extension repo

amwx:
This PR is a redo of #272 (after butchering my local branch) and moves the Avalonia.Ide.CompletionEngine library into this repo so it has better visibility and be brought up to date with newer Avalonia features and updated in general for new/better completion features. 

The completion engine projects have no dependencies on the VS extension and are simply just housed here. The two projects were brought up to netstandard2.0 (from 1.4 I think), if that's an issue let me know, but I don't think it will be. I also have not removed the submodule yet, that's how I butchered the branch of the old PR, but the references for the VS2019 and VS2022 projects are updated to use the local version and not the submodule. I've also only brought the two library projects over.

Also from [this comment](https://github.com/AvaloniaUI/AvaloniaVS/pull/272#discussion_r1040368297), it was copied from the submodule to ensure changes not in master were brought over too

This PR also includes the fixes from #272 I had implemented:
- Added completion for pseudoclasses within Style Selectors
  - This is done by reading the PseudoclassesAttribute on supporting types. It's done through a nasty way, but it works - would be better if we had access to the PseudoclassesAttribute type to directly query that. I've also adjusted some of the completion logic so that pseudoclasses trigger a new completion session so typing "Control:" will list the pseudoclasses, previously because there's no space after 'Control', the engine wouldn't know to refilter the hints and we wouldn't get a list of suggestions since nothing matches "Control:". I've also made the char - NOT end a completion session for pseudoclasses like :focus-within allowing that to be typed without interruption.
ezgif com-gif-maker(1)
- Fixed handling completion commit on '.' key. Currently typing 'Window.Resources', the '.' will cancel the completion session and swallow the '.' character so you'd have to press it twice to continue typing, which is suboptimal. Now the '.' character isn't swallowed and a new completion session is activated after pressing it.
- Fixed the double intellisense popup when adding xml namespaces. I think VS XML service was starting its own and then we were starting one ourselves which was causing the double popups. The existing completion session already had an Avalonia completion set in it, so instead of creating a new session, I've made it just tap into that one.
image

Future TODOs (not in this PR):
- find and list all defined resources instead of using static, hard-coded ones (hopefully, if not will add to future todos) that don't really apply to anything outside the old default theme
  - Ideally with this we need to make sure at least the active document (maybe even the current project) can update the suggestions more frequently without requiring a build (which is how the metadata is currently invalidated) as if you add a resource to the current file, I would expect it to immediately show without having to build the project. External assemblies could probably be cached like normal on build only since they're static unless the ref is explicitly updated
- Fix getting properties in nested style selectors. Nested style selectors without a type `Selector="^:pointerover"` will fail to offer valid suggestions
- Filter out non-public types from suggestions
- Add better context awareness for suggestions, including:
  - filtering the suggestions to only include what makes sense at the given point in the document (this is mostly already how it is, but I think there are a few places where this could be improved, as an example, don't suggest `VerticalContentAlignment` unless it absolutely makes sense to do so (only suggest `VerticalAlignment`)
  - In a Style Selector, we could probably make it only suggest things that make sense given the cursor position, ie don't suggest `/template/` or pseudoclasses at the start of the selector, no pseudoclasses directly after `/template/`, etc
  - In this PR, I've also added some code to tell VS to auto select the first suggestion in the list - previously, it wouldn't usually do this and you had to physically interact (mouse or arrow keys) to get the completion to work. We could probably improve this to guess the best item and select it instead based on context
- Fix editing existing text and accepting a completion
  - currently if you type `ponterover` and try to correct it, adding the `i` will suggest ":pointerover", but accepting that completion will end up with `pointerovernterover` as it just inserts the suggested text and doesn't check anything for replacing

closes #269

cc: @grokys @kekekeks


amwx:
FYI: There's a bunch of errors coming from the Avalonia submodule. They don't affect actually building the project, but I think that might be why CI build is failing here and in my other active PR:

image

maxkatz6:
@amwx build is failing with a newer SDK. Submodule should be updated and local CI SDK as well

maxkatz6 (approved):

maxkatz6:
@amwx I removed that submodule as it won't let solution build. Also moved competition tests to this repo as well.
Updated SDK and Avalonia versions too. Now it's buildable and ready to use.

PR #288, opened 2023-01-18
Remove project templates from the extension

Takoooooo:
We have finished our CLI templates to VS integration and now when people install our CLI templates for VS2022 they would be visible right from the moment they are installed so we don't need to duplicate our template improvements two times now. But i got to leave item templates because item templates installed from dotnet new are currently not visible when you click in VS "Add new item". The next step is to make CLI templates install when people install the VS extension


grokys (approved):

Mrxx99:
@Takoooooo There is also the possibility to include the CLI templates inside Visual Studio (see [here](https://github.com/sayedihashimi/template-sample#how-to-ship-a-template-inside-of-a-visual-studio-extension-vsix)).
That works great and also does not show the templates as duplicates if you have CLI templates installed as well, but it has one drawback: It seems that if CLI templates are installed those are preferred over the VS templates. That does not sound bad at first but when you think about it, it could lead to users to not get updates to newer template vesions if they once installed CLI templates but now rely on VS. Then updating the templates in the VS extension has no effect as the than older CLI templates are still preferred.

Takoooooo:
> @Takoooooo There is also the possibility to include the CLI templates inside Visual Studio (see [here](https://github.com/sayedihashimi/template-sample#how-to-ship-a-template-inside-of-a-visual-studio-extension-vsix)). That works great and also does not show the templates as duplicates if you have CLI templates installed as well, but it has one drawback: It seems that if CLI templates are installed those are preferred over the VS templates. That does not sound bad at first but when you think about it, it could lead to users to not get updates to newer template vesions if they once installed CLI templates but now rely on VS. Then updating the templates in the VS extension has no effect as the than older CLI templates are still preferred. I guess this is a bug in Visual Studio.

I guess it would be more convenient if we had a possibility to install CLI templates when they are not installed or update them when they are installed but this solution should be ok at this point

PR #291, opened 2023-01-25
Include CLI templates in VSIX

Takoooooo:
This PR includes Avalonia CLI templates in our VSIX. But as @Mrxx99 commented it has one drawback. If you have CLI templates they would be preferred over VS extension templates so you will need to update CLI templates first.

### DIFF
--- a/templates/ProjectTemplatesDownloader/ProjectTemplatesDownloader.csproj
+++ b/templates/ProjectTemplatesDownloader/ProjectTemplatesDownloader.csproj
@@ -1,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--This project is needed to download Avalonia CLI templates, we can't download them from main projects because PackageDownload requires SDK-style project
+  and our main projects can't be migrated to SDK-style.-->
+  <PropertyGroup>
+    <TargetFramework>net45</TargetFramework>
+  </PropertyGroup>
+  
+  <ItemGroup>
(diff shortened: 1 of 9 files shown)


maxkatz6:
Do you think we can update pipeline script to download package of specific version instead of hardcoding it on git?

Takoooooo:
@maxkatz6 Should be implemented now,please review

maxkatz6 (approved):

maxkatz6:
Didn't know PackageDownload exist. Looks good though.

templates/ProjectTemplatesDownloader/ProjectTemplatesDownloader.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--This project is needed to download Avalonia CLI templates, we can't download them from main projects because PackageDownload requires SDK-style project
+  and our main projects can't be migrated to SDK-style.-->
+  <PropertyGroup>
+    <TargetFramework>net45</TargetFramework>
```
amwx, line 6: Does this need to be netfx4.5? I can't build this solution anymore as VS complains about not having the 4.5 developer pack installed, and 4.5 isn't supported or available for download anymore (from VS installer or https://aka.ms/msbuild/developerpacks). 4.5.1 and 4.5.2 don't satisfy that requirement either

Takoooooo: no,it doesen't need to be netfx4.5,it could be 4.7.2 or below. It would be helpful if you would make a PR to fix that

PR #297, opened 2023-02-12
Sort completion list

### LINKED ISSUE #296: Better sorting for completion list
Currently the completion list is not sorted and always prompts for the unintended completion. 
For example, when you type `datatem`, you get `DataTemplateExtensions` at the first completion instead of `DataTemplate`.

image

This issue happens even if you type the entire `DataTemplate`:

image

So you will get `DataTemplateExtensions` inserted after you type a `Tab`, which is really annoying.

Maybe AvaloniaVS can sort the completion list before showing up?

hez2010:
Fix #296

### DIFF
--- a/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
+++ b/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
@@ -390,7 +390,7 @@ public class CompletionEngine
         }
 
         if (completions.Count != 0)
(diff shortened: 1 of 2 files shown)


maxkatz6 (approved):

amwx:
I think we might need to revert this and come up with a better solution. Sorting the entire completion list is leading to non ideal ordering of things (placing classes/controls before properties, for example)

image

image

hez2010:
I already have a better solution and will push a PR soon.

Issue #279, opened 2022-12-10
Continued completion engine and intellisense upgrades


amwx:
Just pulling this out of my PR for visibility and tracking, but this is a list of things I marked as TODOs for continuing to upgrade the completion engine and intellisense in VS to support newer Avalonia features and be smarter about the suggestions offered. Some of these are things implemented in the core completion engine, and some are more on the VS intellisense side (or both).

Just to recap what was in #275:
- Improved completion of pseudoclasses, including getting all pseudoclasses from the attribute
- Fixed handling completion on '.' character so it does swallow the '.' in something like `Window.Resources`
- Fixed the double intellisense popup when adding xml namespaces

These are still left:
- [ ] find and list all defined resources instead of using static, hard-coded ones (hopefully, if not will add to future todos) that don't really apply to anything outside the old default theme
  - Ideally with this we need to make sure at least the active document (maybe even the current project) can update the suggestions more frequently without requiring a build (which is how the metadata is currently invalidated) as if you add a resource to the current file, I would expect it to immediately show without having to build the project. External assemblies could probably be cached like normal on build only since they're static unless the ref is explicitly updated
- [ ] Filter out non-public types from suggestions
  - @workgroupengineering Now that the completion engine is in this repo, you should be able to complete your PR (#216) 
- [ ] Fix getting properties in nested style selectors. Nested style selectors without a type Selector="^:pointerover" will fail to offer valid suggestions
- [ ] Add better context awareness for suggestions, including:
  - filtering the suggestions to only include what makes sense at the given point in the document (this is mostly already how it is, but I think there are a few places where this could be improved, as an example, don't suggest VerticalContentAlignment unless it absolutely makes sense to do so (only suggest VerticalAlignment)
  - In a Style Selector, we could probably make it only suggest things that make sense given the cursor position, e.g., don't suggest /template/ or pseudoclasses at the start of the selector, no pseudoclasses directly after /template/, etc
  - In #275, I've also added some code to tell VS to auto select the first suggestion in the list - previously, it wouldn't usually do this and you had to physically interact (mouse or arrow keys) to get the completion to work. We could probably improve this to guess the best item and select it instead based on context
  - Only offering pseudoclasses valid for a particular control
  - Don't suggest properties that are already set on a given control, or within a particular Style block
- Fix editing existing text and accepting a completion
[...]

PR #300, opened 2023-02-20
Better sorting for completions

hez2010:
Addresses https://github.com/AvaloniaUI/AvaloniaVS/pull/297#issuecomment-1436324233

### DIFF
--- a/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
+++ b/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
@@ -390,9 +390,26 @@ public class CompletionEngine
         }
 
         if (completions.Count != 0)
-            return new CompletionSet() { Completions = completions.OrderBy(i => i.Kind).ThenBy(i => i.InsertText).ToList(), StartPosition = curStart };
+            return new CompletionSet() { Completions = SortCompletions(completions), StartPosition = curStart };
 
         return null;
+    }
+
+    private List<Completion> SortCompletions(List<Completion> completions)
+    {
+        // Group the completions based on Kind, and sort the completions for each group
+        var completionGroups = completions
+            .GroupBy(i => i.Kind)
+            .ToDictionary(i => i.Key,
+                i => i.OrderBy(j => j.DisplayText.Length)
+                    .ThenBy(j => j.DisplayText)
+                    .ToList());
+
+        // Resort the groups based on the first completion of each group, and flatten the result to a list
+        return completionGroups.OrderBy(i => i.Value.First().DisplayText.Length)
+            .ThenBy(i => i.Value.First().DisplayText)
+            .SelectMany(i => i.Value)
+            .ToList();
     }
 
     private void ProcessStyleSetter(string setterPropertyName, XmlParser state, List<Completion> completions, string? currentAssemblyName)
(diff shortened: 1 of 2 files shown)


hez2010:
@maxkatz6 It's ready for review. Please take a look.
/cc: @amwx

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -390,11 +390,26 @@ IEnumerable<string> filterNamespaces(Func<string, bool> predicate)
...
+            return new CompletionSet() { Completions = SortCompletions(completions), StartPosition = curStart };
 
         return null;
     }
 
+    private List<Completion> SortCompletions(List<Completion> completions)
+    {
+        // Group the completions based on Kind, and sort the completions for each group
+        var completionGroups = completions
+            .GroupBy(i => i.Kind, (_, compl) =>
```
hez2010, line 402: We may also want to determine the priority of completion Kind.

amwx:
Properties are now first again, but not sorted alphabetically: 
image

image


Selector is still appearing last, this one is more complicated. Typing `<Style `, the space triggers a new completion session to start, which this session lists all the controls first, and Selector is after. Continuing to type `S`, doesn't trigger a new completion session, it just sorts the existing one, which keeps the existing ordering. If you cancel that completion session (escape key) and restart it, it appears as it should.

image


I'm thinking sorting by text length isn't the way to do this (WPF/UWP doesn't do this). There's a couple other minor things that I think aren't so much sorting related as more related to providing more intelligent context-based options (related to #279 which I'm actively working on). Included in this is not suggesting classes like "DataTemplateExtension" and "ControlTemplateResult" that are part of the root problem here. Sorting just alphabetically makes sense, and I would think that should automatically place the list in the order of something like `DataTemplate` -> `DataTemplates` -> `DataTemplateExtension`

hez2010:
I'm considering that we can build a model to predict the most used completion using some machine learning methods to help us prioritize completions in the future.

Mrxx99:
> I'm considering that we can build a model to predict the most used completion using some machine learning methods to help us prioritize completions in the future. 

So you want ro rebuild Intellicode (for XAML)?
Maybe a simpler start would be to add attributes in Avalonia that could be set on properties/events to mark them as common for that control, e.g. Button would mark Content, Command, Click and would also inherit e.g. marked Height/Width and Margin from Control, but could also unmark inherited properties with another attribute or depriorize uncommon properties.
That way not only builtin but also custom controls and controls from libraries could be annotated to improve XAML completion for them.

hez2010:
I've added a simple logic based on my own experience to determine the priority of completions.

tests/CompletionEngineTests/BasicTests.cs
```diff
@@ -135,13 +135,25 @@ public void Extension_DataType_Should_Be_Completed()
...
+        public void Completions_Should_Be_Sorted()
         {
             var compl = GetCompletionsFor("<DataTemplate");
 
-            Assert.Equal(2, compl.Completions.Count);
-            Assert.Equal("DataTemplate", compl.Completions[0].InsertText);
-            Assert.Equal("DataTemplates", compl.Completions[1].InsertText);
+            Assert.Equal(3, compl.Completions.Count);
+            Assert.Equal("DataTemplate", compl.Completions[0].DisplayText);
+            Assert.Equal("DataTemplateExtensions", compl.Completions[1].DisplayText);
```
maxkatz6, line 144: I don't think DataTemplateExtensions should be visible in the completion after https://github.com/AvaloniaUI/AvaloniaVS/pull/302

hez2010: The branch is a bit out-of-date. I've updated the base branch to resolve this.

hez2010:
@amwx Maybe you want to test this PR again? I think this time it should works well.

amwx:
> @amwx Maybe you want to test this PR again? I think this time it should works well.

I did a quick test yesterday I think, and it seemed to work a lot better.

maxkatz6 (approved):

PR #302, opened 2023-02-22
More VS IntelliSense improvements

amwx:
Several improvements, mostly to the VS side of things
- I noticed after the Completion engine refactor, typing something like "Grid." was inserting two periods ("Grid..") for some reason, fixed that
- Updated the logic (even some I had previously done) to improve completion handling to better match the UWP/WPF designers
  - Typing a namespace should no longer take the selected completion option when typing a period. only tab and space can accept a completion while editing a namespace
  - Elsewhere, tab and space can trigger a completion. The tab is always swallowed and isn't entered into the text buffer. A space is only entered into the buffer in a Selector, as it doesn't make sense in most attributes (Background="Red" shouldn't complete to Background="Red ")
- Improved AttachedProperties in Setters. This required a breaking change in the completion engine. In the current version, when the new completion session pop up after the period (like `Grid.`), no options are selected and continuing to type (even something valid like Column or Row for Grid.) would cancel intellisense. The completion engine was providing attached properties with the type name attached ("Grid.Row") which wasn't matching the completion buffer which reset at the period. I changed the completion engine to only supply the property name, which fixes this issue and completion now works entirely for attached properties. 
- Added better support for Markup extension used as XML elements instead of Attributes (StaticResource, OnPlatform, etc)
- Manually filtered out `ControlTemplateResult` and `DataTemplateExtensions` from the type list. They're irrelevant to xaml and seem to be the most commonly noticed. IMO, we really need an attribute for public types to hide them from the designer if the type must be public (EditorBrowsable doesn't work since the type may be useful in C# and it would hide there too)

- Updated to `Completion4`. Not really a change that really matters but uses a more modern VS class. Change removed the IVsImageService from the MEF completion provider as Completion4 uses `ImageMoniker`s instead of `ImageSource`.
- Updated a couple of the icons to better match WPF/UWP, specifically for classes/controls (only difference is that ours is colored, don't know why but a touch of color is kinda nice):
image

and static classes like Brushes now use the enum icon:
image

Still a draft PR as:
- <s>I found an issue taking that last screenshot (typing "Red" then backspacing and changing the color doesn't refresh the completion list)</s>
- There's a couple random errors popping up, one is related to the ErrorTagger, which is irrelevant to the changes made here, and the other is occasionally a new completion session will try to start while another is active. Neither are easily repro'd but if I can figure out why I'll see what I can do.
- Need to fix/add tests

-------

UPDATE:
- Improved completion in markup extensions as whole
- Added full support for the `OnPlatform` and `OnFormFactor` markup extensions, including having the appropriate completions suggested depending on context
- Changed the icon for completion properties when in a DataType/DataContext completion session. The icon is now the same image ID/moniker as the WPF/UWP designer, but it looks a bit different b/c of how old the WPF designer is which isn't using the new icon set (that's my guess at least)

Inline:
ezgif com-video-to-gif

And:
ezgif com-video-to-gif(1)

--------

UPDATE 2:
Changed a couple more icons to match the WPF/UWP: 
- TargetType (and in future types in Selector) use a different icon for class than the one used for tag names ,
- Completions for xmlns use the enum icon
- For these, I changed the `CompletionKind` enum to be essentially a Flags enum so these VS specific kinds don't interfere with other uses

### DIFF
--- a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
@@ -261,7 +261,10 @@ namespace AvaloniaVS.IntelliSense
 
                                 // if in a markup extension, if we skip the entered char, we won't get
                                 // to start a new completion session, so force start it
-                                if (skip)
+                                // The check for '=' in the insertion text ensures we don't always get this
+                                // e.g., {OnPlatform Wind -> {OnPlatform Windows= [New completion session]
+                                // but {OnPlatform Windows=Re -> {OnPlatform Windows=Red [no new session]
+                                if (skip && selected.InsertionText.EndsWith("="))
                                     TriggerCompletion();
                             }
                         }
--- a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
@@ -251,7 +251,7 @@ namespace AvaloniaVS.IntelliSense
                                 for (int i = text.Length - 1; i >= 0; i--)
                                 {
                                     var lineChar = text[i];
-                                    if (char.IsLetter(lineChar) || char.IsNumber(lineChar) || lineChar == ':')
+                                    if (char.IsLetterOrDigit(lineChar) || lineChar == ':')
                                         continue;
 
                                     // any other character than [A-z,0-9,:] is a different part
--- a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
@@ -218,7 +218,35 @@ namespace AvaloniaVS.IntelliSense
                         {
                             // For most xml attributes, swallow the space upon completion
                             // For selector, allow it to go into the buffer
+                            // Also if in a markupextention
                             skip = !(parser.AttributeName?.Equals("Selector") == true);
+
+                            // If we're in a markup extension, only swallow the space if the
+                            // completion isn't on the Markup extension
+                            // i.e., where | is the cursor
+                            // {DynamicResource -> {DynamicResource |
+                            // but {Binding Path= -> {Binding Path=|
+                            // similarly, more embedded things like RelativeSource work the same way
+                            // {Binding path, RelativeSource={RelativeSource -> ...={RelativeSource |
+                            if (parser.AttributeValue?.StartsWith("{") == true)
+                            {
+                                // To determine, we'll walk back the text from the cursor position
+                                // until we hit either something that isn't a character
+                                // If that's a {, we apply the space, otherwise we dont
+                                // Only using the line text (up to cursor) since xaml can't wrap
+                                // Also ignore ':' for namespaces or directives
+                                var text = line.Snapshot.GetText(start, end - start);
+                                for (int i = text.Length - 1; i >= 0; i--)
+                                {
+                                    var lineChar = text[i];
+                                    if (char.IsLetter(lineChar) || char.IsNumber(lineChar) || lineChar == ':')
+                                        continue;
+
+                                    // any other character than [A-z,0-9,:] is a different part
+                                    skip = lineChar != '{';
+                                    break;
+                                }                                
+                            }
                         }
                         else if (c == '\'' || c == '"')
                         {
@@ -237,6 +265,14 @@ namespace AvaloniaVS.IntelliSense
                         {
                             skip = false;
                         }                        
+
+                        // Cases like {Binding Path= result in {Binding Path==
+                        // as the completion includes the '=', if the entered char
+                        // is the same as the last char here, swallow the entered char
+                        if (!skip && selected.InsertionText.EndsWith(c.ToString()))
+                        {
+                            skip = true;
+                    }
                     }
                     else if (state != XmlParser.ParserState.StartElement)
                     {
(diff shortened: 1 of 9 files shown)


CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -240,7 +241,24 @@ public IEnumerable<string> FilterTypeNames(string? prefix, bool withAttachedProp
                     .Select(p => new Completion(p, sameType ? CompletionKind.Property : CompletionKind.AttachedProperty)));
             }
             else
-                completions.AddRange(_helper.FilterTypeNames(tagName).Select(x => new Completion(x, CompletionKind.Class)));
+            {
+                // Special case for `StaticResourceExtension`:
+                // <StaticResource ... />
+                // IIRC we don't support this for DynamicResource
+                // Change the name to "StaticResource" (matches UWP)
+                // TODO: Probably a better way to handle this...
```
maxkatz6, line 249: It's not quite special. 
StaticResourceExtension.ProvideValue returns an object which is then added to the resources dictionary.
DynamicResourceExtension.ProvideValue returns a binding. And actual values are not unwrapped from the bindings in the resources.

maxkatz6: Technically, you can use OnPlatformExtension as well, which would return an object from ProvideValue. And it will work too.

maxkatz6: I think you can do this logic for any markup extension. I.e. a class that has a non-static ProvideValue method (that returns something and might or might not accept single IServiceProvider argument).

amwx: I'll rework it this to work more generally. I also forgot about OnPlatformExtention, which ironically was the other extension I wanted to do this for.

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -484,7 +554,16 @@ public IEnumerable<string> FilterHintValues(MetadataType type, string? entered,
...
+            // Don't filter values here by 'StartsWith' (old behavior), provide all the hints,
+            // For VS, Intellisense will filter the results for us, other users of the completion
+            // engine (outside of VS) will need to filter later
+            // Otherwise, in VS, it's impossible to get the full list for something like brushes:
+            // Background="Red" -> Background="B", will only populate with the 'B' brushes and hitting
+            // backspace after that will keep the 'B' brushes only instead of showing the whole list
+            // WPF/UWP loads the full list of brushes and highlights starting at the B and then
+            // filters the list down from there - otherwise its difficult to keep the completion list
+            // and see all choices if making edits
+            foreach (var v in type.HintValues)
```
amwx, line 566: - I only turned off the filtering (.StartsWith) here, not above with the XamlContextHintValuesFunc - not entirely sure what that does

My thought here is though is that the CompletionEngine shouldn't be doing this filtering anyway, and that should be done in the IDE (this is done in quite a few places in here). At least in VS, intellisense completions handle the filtering and if the CompletionEngine only serves a portion of all potential candidates, VS can't properly recover if you backspace through to correct or see new options. But, I'm not sure where else the CompletionEngine is used and how it would be used so I don't want to make this change without at least a discussion first (also outside this PRs scope). At the very least, a bool flag could be used to opt out for VS but allow in other cases (though that might be difficult with all the linq)
However, this only seems to really matter here for type hints (like static Brushes list).

AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
```diff
@@ -78,7 +93,7 @@ public int Exec(ref Guid pguidCmdGroup, uint nCmdID, uint nCmdexecopt, IntPtr pv
                 {
                     return VSConstants.S_OK;
                 }
-
+
```
workgroupengineering, line 96: nit

AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
```diff
@@ -155,47 +149,206 @@ private bool HandleSessionUpdate(char c)
...
                         {
-                            TriggerCompletion();
+                            skip = false;
                         }
 
-                        return skip;
+                        // Cases like {Binding Path= result in {Binding Path==
+                        // as the completion includes the '=', if the entered char
+                        // is the same as the last char here, swallow the entered char
+                        if (!skip && selected.InsertionText.EndsWith(c.ToString()))
```
workgroupengineering, line 292:
```suggestion
                        if (!skip && selected.InsertionText?.Length > 0 && selected.InsertionText[selected.InsertionText.Length-1] == c)
```

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/Completion.cs
```diff
@@ -2,22 +2,33 @@
 
 public enum CompletionKind
```
maxkatz6, line 3: It should have Flags attribute here, if it's used as flags enum now.

Optionally, renamed to CompletionKinds, but not necessary.

amwx: Added [Flags]. Renaming can be done later, if needed.

maxkatz6 (approved):

PR #303, opened 2023-02-23
feat: Selector autocompletate

workgroupengineering:
Selector Autocompletate

Allow autocompletate for selector:

- Property
- Type 
- Namespace 
- Attached Property
- Name
- Property and Attached Property value (Nullable<T>)

### DIFF
--- a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 using Avalonia.Ide.CompletionEngine;
@@ -103,8 +101,6 @@ namespace AvaloniaVS.IntelliSense
 
         private bool HandleSessionStart(char c)
         {
-            System.Diagnostics.Debug.WriteLine($"HandleSessionStart({((int)c):0x}{(char.IsWhiteSpace(c) ? ' ' : c)})", "Session");
-
             // If the pressed key is a key that can start a completion session.
             if (CompletionEngine.ShouldTriggerCompletionListOn(c) || c == '\a')
             {
@@ -135,7 +131,6 @@ namespace AvaloniaVS.IntelliSense
 
         private bool HandleSessionUpdate(char c)
         {
-            System.Diagnostics.Debug.WriteLine($"HandleSessionUpdate({((int)c):0x}{(char.IsWhiteSpace(c) ? ' ' : c)})", "Session");
             // Update the filter if there is a deletion.
             if (c == '\b')
             {
@@ -157,7 +152,6 @@ namespace AvaloniaVS.IntelliSense
             var start = line.Start;
             var end = Math.Min(line.End, _textView.Caret.Position.BufferPosition);
 
-            System.Diagnostics.Debug.WriteLine($"HandleSessionCompletion(({(char.IsWhiteSpace(c) ? ' ' : c)}))", "Session");
             // Adding a xmlns is special-cased here because we don't want '.' triggering
             // a completion, which can complete on the wrong value
             // So we only trigger on ' ' or '\t', and swallow that so it doesn't get 
@@ -403,28 +397,17 @@ namespace AvaloniaVS.IntelliSense
             return false;
         }
 
-        private bool TriggerCompletion([CallerMemberName] string memberName = "",
-            [CallerFilePath] string fileName = "",
-            [CallerLineNumber] int lineNumber = 0)
+        private bool TriggerCompletion()
         {
-
-            System.Diagnostics.Debug.WriteLine($"TriggerCompletion Call by {memberName} at {lineNumber}", "Session");
-            System.Diagnostics.Debug.WriteLine($"TriggerCompletion Start", "Session");
-
             // The caret must be in a non-projection location.
             var caretPoint = _textView.Caret.Position.Point.GetPoint(
                 x => (!x.ContentType.IsOfType("projection")),
                 PositionAffinity.Predecessor);
 
-            System.Diagnostics.Debug.WriteLine($"TriggerCompletion caretPoint{caretPoint}", "Session");
-
             if (!caretPoint.HasValue)
             {
                 return false;
             }
-
-
-            System.Diagnostics.Debug.WriteLine($"TriggerCompletion Char = {caretPoint.Value.GetChar()}", "Session");
 
             // When adding an xmlns definition, we were getting 2 intellisense popups because (I think)
             // the VS XML intellisense handler was popping one up and then we are creating our own session
@@ -455,8 +438,6 @@ namespace AvaloniaVS.IntelliSense
                         break;
                 }
             }
-
-            System.Diagnostics.Debug.WriteLine($"TriggerCompletion start session {caretPoint.Value.Position}", "Session");
 
             _session = existingSession ?? _completionBroker.CreateCompletionSession(
                 _textView,
--- a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
+++ b/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Runtime.InteropServices;
-using System.Threading;
 using Avalonia.Ide.CompletionEngine;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Language.Intellisense;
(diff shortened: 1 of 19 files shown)


workgroupengineering:
if anyone has time can try it and report any problems. In the tests I'm doing at the moment I haven't found any.

workgroupengineering:
wait merge #302

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/Completion.cs
```diff
@@ -31,14 +31,21 @@ public enum CompletionKind
     /// <summary>
     /// xmlns list in visual studio (uses enum icon instead of namespace icon)
     /// </summary>
-    VS_XMLNS = 0x800
+    VS_XMLNS = 0x800,
+    
+    Selector,
```
amwx, line 36: Make sure to give this a value so it doesn't conflict (flags enum now)

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -856,9 +879,259 @@ public static bool ShouldTriggerCompletionListOn(char typedChar)
...
+                        completions.Add(new Completion(".", CompletionKind.Selector));
+                        completions.Add(new Completion("#", CompletionKind.Selector));
+                        completions.Add(new Completion(":is()", ":is(", CompletionKind.Selector));
+                        completions.Add(new Completion(":not()", ":not(", CompletionKind.Selector));
+                        completions.Add(new Completion(":nth-child()", ":nth-child(", CompletionKind.Selector));
+                        completions.Add(new Completion(":nth-last-child()", ":nth-last-child(", CompletionKind.Selector));
+                        completions.Add(new Completion("tempalte", "/template/", CompletionKind.Selector));
+                        var types = _helper.FilterTypes(default)
+                            .Where(t => t.Value.IsAvaloniaObjectType || t.Value.HasAttachedProperties)
+                            .Select(t => new Completion(t.Value.Name.Replace(":", "|"), CompletionKind.Class));
```
amwx, line 1053: Same here for types

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -856,9 +879,259 @@ public static bool ShouldTriggerCompletionListOn(char typedChar)
...
+                    {
+                        completions.AddRange(type.PseudoClasses.Select(v => new Completion(v, CompletionKind.Selector)));
+                    }
+                    parsered = parser.LastParseredPosition;
+                }
+                break;
+            case SelectorStatment.Name:
+                {
+                    if (fullText is not null)
+                    {
```
amwx, line 913: I have some reservations on this...
From my understanding, you're just scanning the entire document for (x:)Name="..." to populate a list of names. I'm not sure I fully like this. Ideally, this needs to be context aware.

That is, if we're in a ControlTheme, it should only list items within the template. If we're writing a general selector (before the `/template/`), then its fine to list elements in the xaml body that have a name. If we're beyond the `/template/`, we should only list items in the template. That get complicated though if the template isn't in the active document. Some of this can be accomplished with the `TemplatePartAttribute`, for the rest, we probably need the Xaml compiler to get the namescope and all available names, which is a huge task.

To be clear, I don't think this is blocking issue, but it needs some thought. It might be confusing if we suggest a name that isn't relevant to what the user is trying to do (which then has no effect)

AvaloniaVS.Shared/IntelliSense/XamlCompletionSource.cs
```diff
@@ -61,7 +61,7 @@ public void AugmentCompletionSession(ICompletionSession session, IList<Completio
                         applicableTo,
                         XamlCompletion.Create(completions.Completions),
                         null));
-
+                    System.Threading.Thread.Sleep(500);
```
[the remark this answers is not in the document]
workgroupengineering: Good catch, that was just for debugging

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
workgroupengineering: I totally agree your think. But in this moment is more simple solution. We will be able to do when XamlC will be fault tolerance.

workgroupengineering:
@amwx  I finished. You can download the artifact from [here](https://dev.azure.com/AvaloniaUI/aa84306f-2981-47b9-8206-edb3bed6250d/_apis/build/builds/30967/artifacts?artifactName=publish&api-version=7.0&%24format=zip).

amwx:
Found a few issues:

- After `:not()`, `:is()`, etc, trying to add a name after will only suggest pseudoclasses (same thing if you type '/` to start /template/ without a space, which is allowed by the compiler). Normal pseudoclasses work fine, (e.g., Button:pointerover# will present names)

image

- Names aren't auto selected in the intellisense popup, requiring you to manually select them:
image

What I think is happening here, (line 374 in CompletionEngine), you're calculating the current start for the autocompletion as `textToCursor.Length - delta` where the delta includes the `#`, but all the completion sets are built without the `#` on the `DisplayText` member, which is what VS uses to match typed text. Adjusting the start position to start after the `#` should fix this


- Trying to add names or pseudoclasses after a property doesn't give autocomplete suggestions (hitting space will resume normally):
`Button[Tag=Value]:` or `Button[Tag=Value]#`. 


- I think pseudoclasses should be suggested in cases where its appropriate for types (the operators like `:is()`, `:not()` are the only things suggested after `:` if there's a space). Otherwise something like `Button :pointerover` won't suggest the pseudoclasses (there's no spec that says pseudoclasses, etc have to be right up against the type so spaces are allowed). This also fails in ControlTheme child styles:

image

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -856,9 +879,259 @@ public static bool ShouldTriggerCompletionListOn(char typedChar)
...
 
     public static CompletionKind GetCompletionKindForHintValues(MetadataType type)
         => type.IsEnum ? CompletionKind.Enum : CompletionKind.StaticProperty;
+
+
+    public int? ProcesssSelector(ReadOnlySpan<char> text, XmlParser state, List<Completion> completions, string? currentAssemblyName, string? fullText)
+    {
+        int? parsered = default;
+        var parser = SelectorParser.Parse(text);
+        Debug.WriteLine(parser.Statment);
```
amwx, line 893: Debug statement

CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
```diff
@@ -856,9 +879,259 @@ public static bool ShouldTriggerCompletionListOn(char typedChar)
...
+                                filterName = filterName
+                                    .Where(m => m.Groups["AttribValue"].Value.StartsWith(elementName, StringComparison.OrdinalIgnoreCase));
+                            }
+                            foreach (Match m in filterName)
+                            {
+                                if (m.Success)
+                                {
+                                    parsered = text.Length - (parser.LastParseredPosition - 1);
+                                    var name = m.Groups["AttribValue"].Value;
+                                    completions.Add(new Completion(name, $"#{name}", name, CompletionKind.Class | CompletionKind.TargetTypeClass));
```
amwx, line 932: Names can just be `Class`. Otherwise everything else looks fine

workgroupengineering:
Hi @amwx  [here](https://dev.azure.com/AvaloniaUI/aa84306f-2981-47b9-8206-edb3bed6250d/_apis/build/builds/31843/artifacts?artifactName=publish&api-version=7.0&%24format=zip) there is a new build.
Can you test?
There is lot of debug code, i will remove after.

amwx:
From above:
1- typing something like `Button:not(:disabled)`, closing the parenthesis for :not() will enter an infinite loop & freeze VS. I didn't have the debugger attached so I don't know where

2- Still not fixed (names aren't auto selected)

3- Still not fixed (no suggestions are offered after a property unless you hit space)

4- fixed

workgroupengineering:
> 

Thanks again for taking the time to test.

[Here](https://dev.azure.com/AvaloniaUI/aa84306f-2981-47b9-8206-edb3bed6250d/_apis/build/builds/31850/artifacts?artifactName=publish&api-version=7.0&%24format=zip) you can find the build with fixes.

I added parts name detection.
If you type # after templete it shows you the list of parts, or after the template enter the type and pressing # it displays only the parts of the corresponding type.

amwx:
Latest changes look good to me. Adding support for `TemplatePartAttribute` is nice feature

Item 2 still isn't fixed, but I'll handle that in the future so don't worry about it. Its related to differences between what we send to our completion engine and what VS uses internally and I'm seeing that affect other places too, and in #302 I hinted I wanted to rework that aspect anyway, but that's a bigger project

CompletionEngine/Avalonia.Ide.CompletionEngine/Parsing/SelectorParser.cs
```diff
@@ -0,0 +1,818 @@
...
+    public string? Value =>
+        _context.GetRange(_context.ValueStart, _context.ValueEnd).ToString();
+
+    public string? ElementName =>
+        _context.GetRange(_context.NameStart, _context.NameEnd).ToString();
+
+    public string? FunctionName =>
+        _context.GetRange(_context.FuntcionNameStart, _context.FunctionNameEnd).ToString();
+
+    public SelectorStatment Statment =>
```
amwx, line 259:
```suggestion
    public SelectorStatement Statement =>
```

CompletionEngine/Avalonia.Ide.CompletionEngine/Parsing/SelectorParser.cs
```diff
@@ -0,0 +1,818 @@
...
+        _context.Statment;
+
+    public bool IsError =>
+        _context.IsError;
+
+    public int? LastParseredPosition =>
```
amwx, line 265:
```suggestion
    public int? LastParsedPosition =>
```

CompletionEngine/Avalonia.Ide.CompletionEngine/Parsing/SelectorParser.cs
```diff
@@ -0,0 +1,818 @@
...
+        _context.LastParseredPosition;
+
+    public SelectorStatment PreviusStatment =>
```
amwx, line 268:
```suggestion
    public SelectorStatement PreviousStatement =>
```

amwx (commented):
Mostly just typos in "public" (or relating to public) members. I've only tagged their declaration, the usages will need to be updated too - but VS's renaming refactor should make that easy

AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs
```diff
@@ -190,14 +188,34 @@ private bool HandleSessionCompletion(char c)
...
+                            : new(bufferPos, rof);
+                        System.Threading.Tasks.Task.Factory.StartNew(stateArg =>
+                        {
+                            var span = (SnapshotSpan)stateArg;
+                            _textView.TextBuffer.Replace(span, string.Empty);
+                        }, ss
+                        , CancellationToken.None
+                        , System.Threading.Tasks.TaskCreationOptions.None
+                        , System.Threading.Tasks.TaskScheduler.FromCurrentSynchronizationContext());
+                    }
```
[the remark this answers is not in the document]
workgroupengineering: When complete template part remove prefix (eg: Panel#)

grokys:
Sorry for the delay in approving/merging these PRs, we've not had much time recently to review the VS extension PRs. Will try to resolve that shortly.

maxkatz6 (approved):

Issue #304, opened 2023-02-27
Issue after merge PR #302


workgroupengineering:
When you type new element and then hit the space key it shows more element of expected.

![Autocompletate Issue 01](https://user-images.githubusercontent.com/12531229/221530620-44368e1d-c78b-4fac-a243-e4b3d4a69d37.gif)

## Expected

.01 Property
.02 Attached Property
.03 Only Public Event 
.04 Dicercive

PR #307, opened 2023-02-27
feat: Allows filtering of completion suggestions by member type.

workgroupengineering:
Intellisense Filter
Part of: #302, #304


maxkatz6:
On the video, right after you completed 'Button' part, shouldn't it start suggesting properties and events (combined group?) instead of suggesting other controls? I think that's how WPF did it, could be double checked.

PR #308, opened 2023-02-28
Remove item templates because now they are included in VS through CLI

Takoooooo:


Mrxx99:
I think that only works for VS2022, should they still be removed from VS2019 extension?

Takoooooo:
> I think that only works for VS2022, should they still be removed from VS2019 extension?

I guess you are right,will revert. We are planning to deprecate VS2019 support but currently let's revert that

Takoooooo:
Doesen't work currently,seems that only project templates are getting included currently

Mrxx99:
Hm, maybe it does not work for embedded nuget templates...

Mrxx99:
> We are planning to deprecate VS2019

The current VS2019 extension release is already not working, I can't find any templates inside, and even if it worked, the current templates use file scoped namespaces, which are not supported in .NET6 and VS2019 does not support .NET7. So I think the best approach would be to revert the VS2019 extension to the last release where templates were still included in the extension itself and then deprecated it, so it is at least usable but deprecated

Mrxx99:
> Doesen't work currently,seems that only project templates are getting included currently

For me it works in a sample VSIX, that the item template is included in extension (dotnet new template not installed).
But what I noticed and what maybe let you to that conclusion is, that if you open the add new item dialog, initially all folders are shown, but only when the loading text disappears, the folder for my test item template appeared, so if you are not waiting long enough you may have missed it. But will also try to test the Avalonia extension

maxkatz6:
I would vote to drop templates from VS 2019 too. And ask users to use CLI with legacy IDE. Feels like a reasonable trade off.

Takoooooo:
> I would vote to drop templates from VS 2019 too. And ask users to use CLI with legacy IDE. Feels like a reasonable trade off.

I am bringing back project templates for a while,making a last release of vs2019 and deprecating vs2019 as target i think.

PR #310, opened 2023-03-01
fix: Intellisense sorting

### LINKED ISSUE #304: Issue after merge PR #302

workgroupengineering:
closes #304

### DIFF
--- a/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
+++ b/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs
@@ -90,7 +90,9 @@ public class CompletionEngine
 
         public IEnumerable<string> FilterTypeNames(string? prefix, bool withAttachedPropertiesOrEventsOnly = false, bool markupExtensionsOnly = false, bool staticGettersOnly = false, bool xamlDirectiveOnly = false)
         {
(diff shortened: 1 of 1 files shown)


amwx:
This is the third PR trying to fix this issue

maxkatz6:
@amwx @workgroupengineering @hez2010 how would you recommend dealing with these PRs #300 #307 #310? Which should be merged or not?

The problem is that I don't have access to the Visual Studio to test them all. The biggest difference is grouping in one PR and sorting by completion type in another as it seems. Something that seems to be heavily based on the preference of developer.
From Rider experience, they don't have sorting by alphabet by default, and instead show first namespaces, and then something that looks like popular usage based sorting.

amwx:
I tested #300 the other day and it seems to work fine after the most recent changes. I haven't tested this or #307. The Xaml designer for WPF/UWP sorts everything by:
- Intellicode suggestions (starred items), probably sorted by most likely to be used as determined by intellicode
- All possible options (sorted alphabetically, no grouping by suggestion type)

XAML1

XAML2

The two biggest issues with completion in our designer is: types that are added to the completion system that should be (like DataTemplateExtensions or ControlTemplateResult), and certain suggestions being out of order making it harder to take a completion, like `HorizontalContentAlignment` appearing before `HorizontalAlignment`. Sorting takes care of the latter, the irrelevant types are a bit more involved.
IMO, I'd take #300, it seems to be the simplest solution and was done first.

PR #312, opened 2023-03-01
feat: Allow AvaloniaVS defered load

### LINKED ISSUE #311: Load Extension only Avalonia Project
## Current behavior

When any solution is opened, the AvaloniaVS extension is loaded.

## Expected
AvaloniaVS extension is loaded when add reference at Avalonia nuget package

workgroupengineering:
Allow loading the AvaloniaVS extension when referencing the Avalonia nuget package or add ProjectCapability Avalonia to project

[Doc](https://github.com/microsoft/VSProjectSystem/blob/2a8735daf164b0c2f93ab968357dfd278200595d/doc/overview/dynamicCapabilities.md)

### Example to Add ProjectCapability  to project
```xml
  <ItemGroup>
    <ProjectCapability Include="Avalonia"/>
  </ItemGroup>
```

Fixes #311
Depend of  AvaloniaUI/Avalonia#10523
Part of #279

### DIFF
--- a/AvaloniaVS.Shared/AvaloniaPackage.cs
+++ b/AvaloniaVS.Shared/AvaloniaPackage.cs
@@ -16,7 +16,6 @@ namespace AvaloniaVS
     [Guid(Constants.PackageGuidString)]
     [InstalledProductRegistration("#110", "#112", "1.0", IconResourceID = 400)]
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
-    [ProvideAutoLoad(VSConstants.UICONTEXT.SolutionExistsAndFullyLoaded_string, PackageAutoLoadFlags.BackgroundLoad)]
     [ProvideEditorExtension(
         typeof(EditorFactory),
         "." + Constants.axaml,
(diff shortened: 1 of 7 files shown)


Mrxx99:
Will the extension than still work for projects with Avalonia versions where this is not set?

maxkatz6:
Yes. It needs to be confirmed. 
Avalonia PR should be backported to 0.10.19 too.

workgroupengineering:
Yes work. For version <= 0.10.18 you shold add following to csproj

```xml
  <ItemGroup>
    <ProjectCapability Include="Avalonia"/>
  </ItemGroup>
```

you may think, at first, of adding these lines to the templates.

Mrxx99:
But couldn't that still lead to many users thinking the extension is broken after updating it when using it with an existing project? Maybe that is a tradeoff that is okay and maybe a pinned issue in the repository would be enough but that change still has the potential to flood Telegram and even the main Avalonia repo with "bug reports". And the usecase that you clone another project and you first have to set that property before the extension works, needs to be considered too.

I am not against this change per se, just raising some points to consider.

maxkatz6 (approved):

maxkatz6:
@Mrxx99 @workgroupengineering ProjectCapability was included with 0.10.20+ and 11.0 versions. As well as in newer templates.
It shouldn't be an issue for new users, but only for users who installed a new extension with an old Avalonia.
Meaning, at this point, should be safe to merge this PR and help users of old versions.

PR #318, opened 2023-03-11
feat: Avoid dependencies mix match issue

workgroupengineering:
This attribute registers a path that should be probed for candidate assemblies at assembly load time.


maxkatz6:
@workgroupengineering after last merge commit there is no changes in this PR

PR #333, opened 2023-06-06
Install Avalonia TemplateStudio extension when installing AvaloniaVS

Takoooooo:
From now when you install AvaloniaVS for VS 2022 you will get this dialog
image
And you will get TS for Avalonia installed with AvaloniaVS

### DIFF
--- a/AvaloniaVS.VS2022/AvaloniaVS.VS2022.csproj
+++ b/AvaloniaVS.VS2022/AvaloniaVS.VS2022.csproj
@@ -53,6 +53,9 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <Content Include="Extension.vsext">
+      <IncludeInVSIX>true</IncludeInVSIX>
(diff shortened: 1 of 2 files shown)


maxkatz6 (approved):

Mrxx99:
This does not install the TemplateStudio if you update the Avalonia extension.
That means if you don't have CLI templates installed and update the Avalonia extension, there is no Avalonia template to be found in Visual Studio as the CLI templates where also removed from this extension. Would it not make sense to still keep them? At least I would still like them to be included as it is faster if you e.g. only want desktop.

Also for the problem that the TemplateStudio is not installed automatically on update, if it is not possible to integrate that maybe the extension could show a Notification to warn about missing TemplateStudio.
This  notification should also work with defered loading.

Mrxx99:
@Takoooooo Also now it is not possible to create an Avalonia app with version 0.10.x, as the TemplateStudio does not support this. Another reason to add the CLI templates back to the VS extension. They could just live together in peace ;)

PR #336, opened 2023-06-09
Fix intellisense attached property type mix match

workgroupengineering:
To simulate the issue, install the latest build from the repository and try adding any `Grid.Column` element, you can see that Intellisense proposes `True/False`.

The behavior depends on this fragment

https://github.com/AvaloniaUI/AvaloniaVS/blob/20556a572c1a8a0bc89da34b2e6f6dc4057402e9/CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs#L270-L286

which returns the first static methods initialed by `Set` or `Get` instead of checking the full name.

This lines:

https://github.com/workgroupengineering/AvaloniaVS/blob/087973fd16142f327131f0107fb1bce73c0bd8e5/CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs#L557-L566

is adding for check currect retutn type in test.

### DIFF
--- a/CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs
+++ b/CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs
@@ -532,7 +532,7 @@ public static class MetadataConverter
 
     private static void PreProcessTypes(Dictionary<string, MetadataType> types, Metadata metadata)
     {
-        MetadataType xDataType, xCompiledBindings, boolType, typeType;
+        MetadataType xDataType, xCompiledBindings, boolType, typeType, int32Type;
         var toAdd = new List<MetadataType>
         {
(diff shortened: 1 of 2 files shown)


CompletionEngine/Avalonia.Ide.CompletionEngine/AssemblyMetadata/MetadataConverter.cs
```diff
@@ -267,15 +267,18 @@ public static Metadata ConvertMetadata(IMetadataReaderSession provider)
                                 IMethodInformation? setMethod = null;
                                 IMethodInformation? getMethod = null;
 
+                                var setMenthdName = $"Set{name}";
```
Sorien, line 270: typo `Menthd` -> `Method`?

maxkatz6 (approved):

PR #382, opened 2023-08-31
fix: ArgumentOutOfRangeException

workgroupengineering:
fix ArgumentOutOfRangeException causing from my PR #375.

I forgot to add the 'Extension' postfix causing `ArgumentOutOfRangeException` in the following lines:

https://github.com/AvaloniaUI/AvaloniaVS/blob/b25179d7354c2ec951a622fbc0ab9d2ff78e4606/CompletionEngine/Avalonia.Ide.CompletionEngine/Completion/CompletionEngine.cs#L282-L286


maxkatz6:
Wouldn't it be better if CompletionEngine code could understand that not all markup extensions names end with "Extension"?

maxkatz6 (approved):

PR #388, opened 2023-09-11
Add support for event completions

### LINKED ISSUE #317: [Feature Request]: generate code behind method if an Event is defined in the XAML
In WPF, if you type this XAML

```xml
<Button Click="Button_Click" />
```

a new method will be added in code behind (if not present) that matches to the reqiuered argument args

```cs
private void Button_Click (object sender, RoutedEventArgs e) 
{
   ...
}
```

is it possible to add this to the Avalonia extension as well? 

Note: Rider supports this, but not everyone can switch to Rider. 

Related: https://github.com/AvaloniaUI/Avalonia/discussions/10621 

/cc @giziben

Takoooooo:
Now when you will type the event name the AvaloniaVS extension will generate an event handler
Анимация1
Note: Listing available handlers in the completion list is not supported yet. I can't find a way to make it work
Fixes #317


kekekeks (approved):
- extract new code for HandleCompletion to a separate method
- check if 4.7.0 is compatible with VS2017

Otherwise lgtm

Takoooooo:
1) Fixed
2) It doesen't even work with VS2019 currently,latest version which should work with vs2017 should be 2.10 according to https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md  .I am planning to fix VSIX for vs2017 first and update this PR later

Takoooooo:
Now when i've investigated the VS2017 support the plan is to only support VS2019\VS2022. We are only doing one release with VS2017 support and then deprecating it. From now the extension works with this PR for both VS2022/VS2019

PR #407, opened 2023-10-18
feat(Previewer):  Show last view on error.

workgroupengineering:
## Before

Before

## After

after


AvaloniaVS.Shared/Views/AvaloniaPreviewer.xaml.cs
```diff
@@ -138,6 +138,15 @@ private async void Update(object sender, EventArgs e)
 
         private void Update(BitmapSource bitmap)
         {
+            if (Process is null)
+            {
+                return;
+            }
+            if (bitmap is null && _lastBitmap.TryGetTarget(out bitmap))
```
prashantvc, line 145: Could you remove this empty condition?

prashantvc (changes requested):
Requested for one change

maxkatz6 (approved):

Issue #357, opened 2023-07-11
Incorrect auto-completion of types with no public constructors


maxkatz6:
### Describe the bug

XAML compiler only accepts these options to create an instance of the XAML child:
1. Public constructor
2. Constructor with single IServiceProvider argument
3. x:Arguments syntax https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/xarguments-directive (we do not support x:FactoryMethod  though)

### To Reproduce

1. Start typing
2. Scroll suggestions until there is "Func"
3. Accept the suggestion
4. Now we have an invalid XAML that can't create an instance of the FuncControlTemplate

### AvaloniaVS plugin version

0.10.18.4

### Relevant log output

```shell
-
```

### Additional context

![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/3163374/1123089c-18df-4c70-8abb-5b7bf3e96767)

maxkatz6:
Once again I would expect one of these solutions:
1. Either hide any type without public constructor (or single IServiceProvider) until user starts typing this type name (i.e. expecting Func... for example)
2. Or inject full node with x:Arguments.
3. Or a combination of the two above.

PR #410, opened 2023-10-28
feat: auto-completition with generics

### LINKED ISSUE #356: Incorrect auto-completition with generics
### Describe the bug

Auto-completion adds generics types, which technically can be used from XAML, but in a very different syntax.
Two options I see:
1. Either ignore generic types from auto-completion at all.
2. Or transform it to a XAML syntax with x:TypeArguments and move cursor to the inside of the x:TypeArguments directive.

Haven't checked how it works in Rider.

### To Reproduce

1. Start typing any type that has generics
2. See auto-completion adds "`1" in the end
3. Accept auto-completion, it will also add "`1" in the XAML

### AvaloniaVS plugin version

0.10.18.4

### Avalonia version

_No response_

### Visual Studio version

_No response_

### Relevant log output

```shell
-
```

### Additional context

image

workgroupengineering:
This only resolve #356. After thi PR merger will work on  #357

## Before

Before

## After

after

Depends on PR 409
Fixes #356
Part of https://github.com/AvaloniaUI/AvaloniaVSCode/issues/8


prashantvc (approved):
Thanks for the changes

PR #445, opened 2024-01-31
fix: Intellisense does not show internal members

workgroupengineering:
## Expected behavior

Internal members will be display.

## Cuttent behavior

Internal members are displayed randomly.

## What PR does

The current code assumes that the TargetAssembly is the first assembly in the `$(IntermediateOutputPath)/Avalonia/references` file, but this is not always the case. The current PR ensures that, regardless of the order of the present assemblies `$(IntermediateOutputPath)/Avalonia/references`, the TargetAssembly is always first.


CompletionEngine/Avalonia.Ide.CompletionEngine.DnlibMetadataProvider/DnlibMetadataProvider.cs
```diff
@@ -9,10 +9,27 @@ namespace Avalonia.Ide.CompletionEngine.DnlibMetadataProvider;
...
+    {
+
+    }
+
+    public DnlibMetadataProvider(string? xamlAssemblyPath)
+    {
+        _xamlAssemblyPath = xamlAssemblyPath;
+    }
 
     public IMetadataReaderSession GetMetadata(IEnumerable<string> paths)
```
maxkatz6, line 24: Why not include xamlAssemblyPath in the paths enumerable instead? I.e. keep it outside of metadata provider responsibility.

maxkatz6 (approved):

PR #453, opened 2024-03-13
Remove templates from the project and install CLI templates for users who don't have them

Takoooooo:
@maxkatz6 May you please test it on your machine? On mine everything works fine, but it's already a bit late for me..


maxkatz6 (approved):

maxkatz6:
Installed extension from the package. Works fine, it installs missing templates. But template studio needs to be updated independently.

PR #458, opened 2024-03-19
Fix typos; Add concat-like `Metadata.AddMetadata(Metadata)`; Add in-line docs for `AvaloniaCompilationAssemblyProvider` usage

BinToss:
# TODO
- [x] update references in AvaloniaVS extensions; Remove `Obsolete` symbols from AvaloniaVS.Shared
- [x] Investigate azure-pipelines build errors I'm unable to reproduce.
    `AvaloniaVS.Shared\Models\ProjectOutputInfo.cs(30,10): Error CS0246: The type or namespace name 'ObsoleteAttribute' could not be found (are you missing a using directive or an assembly reference?)`
    `AvaloniaVS.Shared\Models\ProjectOutputInfo.cs(30,10): Error CS0246: The type or namespace name 'Obsolete' could not be found (are you missing a using directive or an assembly reference?)`

# Features/Enhancements
- Added instance method `Avalonia.Ide.CompletionEngine.Metadata.AddMetadata(Metadata metadata)` to merge dictionaries from one Metadata into another. Keys are only added if they are not already present. Existing keys' values remain unchanged.
    This can be used to add metadata from various build contexts e.g. TargetFrameworks (build for multiple frameworks) and if-else compiler statements. Some TODO comments indicate potential improvements. How do we indicate which context a metadatum came from?

# Miscellaneous Changes
- Add xmldocs for `AvaloniaCompilationAssemblyProvider` `ctor` and `GetAssemblies()`. The API's usage misunderstood in AvaloniaVSCode's AvaloniaLanguageServer, causing builds to fail.
- Typos in `AvaloniaVS.Shared` and `Avalonia.Ide.CompletionEngine` corrected where understood
- Where affected by typos, Public API members' compatibility is maintained via aliases-like methods and properties ('=>' to correct definition). These aliases have been given the `Obsolete` annotation. Method aliases are aggressively in-lined.

### DIFF
--- a/AvaloniaVS.Shared/Models/ProjectOutputInfo.cs
+++ b/AvaloniaVS.Shared/Models/ProjectOutputInfo.cs
@@ -27,8 +27,6 @@ namespace AvaloniaVS.Models
         /// </summary>
         public string RuntimeIdentifier { get; }
 
-        [Obsolete($"Member name mispelled. Use {nameof(TargetPlatformIdentifier)} instead.")]
-        public string TargetPlatfromIdentifier => TargetPlatformIdentifier;
         public string TargetPlatformIdentifier { get; }
 
         /// <summary>
--- a/AvaloniaVS.Shared/Models/ProjectOutputInfo.cs
+++ b/AvaloniaVS.Shared/Models/ProjectOutputInfo.cs
@@ -1,4 +1,4 @@
-﻿using AvaloniaVS.Utils;
+using AvaloniaVS.Utils;
 
 namespace AvaloniaVS.Models
 {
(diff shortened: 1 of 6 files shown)


AvaloniaVS.Shared/Models/ProjectOutputInfo.cs
```diff
@@ -27,7 +27,9 @@ public class ProjectOutputInfo
         /// </summary>
         public string RuntimeIdentifier { get; }
 
-        public string TargetPlatfromIdentifier { get; }
+        [Obsolete($"Member name mispelled. Use {nameof(TargetPlatformIdentifier)} instead.")]
```
maxkatz6, line 30: It's not part of the public API, do we need to keep old members?

BinToss: I used the following rules while refactoring:
- Private or local? Rename.
- If `InternalsVisibleTo`, a new symbol with correct spelling is added to the Internal API. Else, symbols are renamed.
- Public API receives a new symbol with the correct spelling.

maxkatz6: AvaloniaVS is not a library, these APIs are used only in the extension itself.

BinToss: Understood. I'll update references in the extension projects and remove the old symbols.

maxkatz6 (approved):
lgtm

PR #469, opened 2024-04-16
fix: Random NRE

workgroupengineering:
Random Avalonia VS throws a NRE, it depends on the following code

https://github.com/AvaloniaUI/AvaloniaVS/blob/a297a692b45f16004b462a5f76fd423efd82cb8a/AvaloniaVS.Shared/IntelliSense/XamlCompletionCommandHandler.cs#L470-L474

To avoid the issue , I copy the session reference into the local variable before using the session.


AvaloniaVS.Shared/EnumerableExtesions.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+
+namespace AvaloniaVS
+{
+    static class EnumerableExtesions
+    {
+        public static TSource FirstOrDefault<TSource, TArg>(this IEnumerable<TSource> source, Func<TSource, TArg, bool> predicate, TArg arg)
```
maxkatz6, line 8: As previously, this extension method probably more slows down than reduces allocation :)
Just from the standpoint that build in LINQ gets more optimized than just MoveNext.

AvaloniaVS.Shared/EnumerableExtesions.cs
```diff
@@ -0,0 +1,25 @@
...
```
maxkatz6, line 8: Sorry, I meant without extension method at all. 
https://github.com/dotnet/runtime/blob/3cea1eeb6c04ebb7e81d47393b96ad51b66b8d1f/src/libraries/System.Linq/src/System/Linq/First.cs#L65-L103

So I recommend just to keep normal FirstOrDefault, even with a small allocation of a closure.

maxkatz6 (approved):

Issue #472, opened 2024-05-06
"No executable found" when integrating Microsoft.ML.OnnxRuntime.DirectML


llfab:
### Describe the bug

When I have integrated a reference to DirectML using 
`<PackageReference Include="Microsoft.ML.OnnxRuntime.DirectML" Version="1.17.3" Condition="'$(Configuration)' != 'Debug'" />`
the previewer fails with "No executable found". 
![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/31031996/a759c850-f4c0-4254-9040-22bd7f033d06)

The Log is as follows:
![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/31031996/f2d3128a-eb62-41f8-8dc9-eef7f02dd8ba)

When I switch to Debug (where DirectML is deactivated), close the last xaml document and re-open the xaml document, the previewer works with the following output:
![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/31031996/9b185b86-6233-415d-bc0f-a66a186ad738)

This may be due to DirectML being Windows only. Potentially, that changes the Runtime Identifier or TargetPlatformIdentifier under the hood causing the previewer execution detection to fail.

### To Reproduce

As above integrate DirectML, rebuild and re-open xaml file

### AvaloniaVS plugin version

11.5

### Avalonia version

11.0.10

### Visual Studio version

17.9.6

### Relevant log output

```shell
08:30:11.667 [Information]  Stopping previewer process
08:30:13.381 [Error]  No executable found
08:30:17.100 [Information] 0 Stopping previewer process
08:30:17.186 [Error]  No executable found
08:30:19.955 [Information] 0 Stopping previewer process
08:30:20.032 [Error]  No executable found
```

llfab:
Debugging the VS extension shows that when DirectML is referenced the project is not detected as valid executable project:
![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/31031996/7ae0f86a-7d3a-4158-9a80-6eafa62ae419)

because the .IsExecutable is false:
![image](https://github.com/AvaloniaUI/AvaloniaVS/assets/31031996/b60a4c54-1e18-4dc2-9e9a-57386c33c24f)

This is not the case if DirectML is not referenced.

llfab:
Double checking the same project with the Rider extension works both in Debug and Release, hence, with DirectML referenced and not referenced.

llfab:
This works. 
[...]

PR #473, opened 2024-05-06
Do not require the "SelectedTarget" to be an executable project

llfab:
In some occasions the target project is actually not an executable project where `ProjectInfo.IsExecutable` return false. In such a case previewer runs into an error path "no executable found". 

This should fix: https://github.com/AvaloniaUI/AvaloniaVS/issues/472

It seems that the Rider integration also does not check for such a condition.


llfab:
Actually, this is not the solution to the problem. It seems that there can be nuget packages that would cause the exe project to loose its Properties, such that the extension does not recognize it as .IsExecutable

PR #493, opened 2024-08-21
fix: Fit to Width

workgroupengineering:
fix Fit to Width retrency issue


AvaloniaVS.Shared/Views/AvaloniaDesigner.xaml.cs
```diff
@@ -821,11 +819,10 @@ private void UpdateXaml(string xaml)
 
         private void UpdateScaling(double scaling)
```
maxkatz6, line 820: The same method is used in different contexts. Are you sure it's safe to ignore DPI in all of them?

PR #494, opened 2024-08-26
Fixed jagged lines on designer preview

codecat:
Sometimes, on the previewer window, we can see jagged lines because the previewer's size has half pixels.

For example, before this change:

image

After this change:

image


maxkatz6 (approved):
Thank you!

PR #495, opened 2024-08-26
Added option to swap editor and preview panes by default

codecat:
This adds an option to swap the preview and editor panes by default.

image


maxkatz6 (approved):
LGTM, thank you!

PR #497, opened 2024-09-23
Fixing the sizing of the grid in the options view

Keboo:
I run a larger than normal environment font size inside of Visual Studio as I do lots of presentations and want things easily visible. 

I have included before and after screen shots for clarity (grid lines enabled just for screenshots; they are not enabled in the change).

Before:
image

After:
image


maxkatz6 (approved):
Thanks

PR #528, opened 2025-08-14
Filtering autocompletion properties with Contains instead of StartsWith

Thor181:
Previously filtering of the completion was done with StartsWith. This is not problem if you know what do you want find, but if you are a beginner, you may not know whether an element has a property or not and you want to find it.
Before:
Before

After:
After


Thor181:
The tests fail because they are specific to StartsWith. Do the tests need to be rewritten to accommodate the changes?

